PR #57, opened 2026-06-06
fix: get-playlists --playlist-ids crashes with map is not a function

### LINKED ISSUE #56: bug: get-playlists --playlist-ids crashes with "playlistIds.map is not a function"
## Bug Description

`get-playlists --playlist-ids` crashes immediately after fetching begins with a runtime TypeError.

## Steps to Reproduce

```bash
staqan-yt get-playlists --playlist-ids <id1> <id2>
```

## Actual Behavior

```
- Fetching playlist information...
✖ Failed to fetch playlist information

✗ playlistIds.map is not a function
```

## Expected Behavior

Returns metadata for the specified playlists, same as `get-playlist` does for a single ID.

## Root Cause (suspected)

Commander likely delivers `--playlist-ids` as a string when only one ID is passed (or possibly always), causing the downstream code to call `.map()` on a string instead of an array. Compare with how `get-videos --video-ids` handles its variadic flag — the fix is likely the same pattern.

## Notes

- Confirmed present in both the current Homebrew release and the `fix/security-vulnerabilities` branch — this is a pre-existing bug, not a regression
- `get-playlist` (singular) works correctly

prog893:
## Summary

- Commander delivers variadic `--playlist-ids` values in `options.playlistIds`, not as a positional argument
- The function signature had `playlistIds` as the first param, so it was receiving the options object and calling `.map()` on it
- Fixed by reading `options.playlistIds` instead, matching the pattern used by `get-videos`

## Test plan

- [x] `get-playlists --playlist-ids <id1> <id2>` no longer crashes
- [x] Returns correct playlist metadata
- [x] `npm run type-check` clean
- [x] `npm run build` clean

Closes #56

## Summary by CodeRabbit

* **Refactor**
  * Streamlined the playlists command interface for consistency: the command now accepts a single consolidated options object, simplifying how playlist selections and output preferences are provided and processed.

### DIFF
--- a/commands/get-playlists.ts
+++ b/commands/get-playlists.ts
@@ -5,10 +5,13 @@ import { getOutputFormat } from '../lib/config';
 import { formatJson, formatTable, formatCsv } from '../lib/formatters';
 import { OutputOption, VerboseOption } from '../types';
 
-async function getPlaylistsCommand(playlistIds: string[], options: OutputOption & VerboseOption): Promise<void> {
+interface PlaylistIdsOption { playlistIds: string[]; }
+
+async function getPlaylistsCommand(options: PlaylistIdsOption & OutputOption & VerboseOption): Promise<void> {
   initCommand(options);
 
   await withSpinner('Fetching playlist information...', 'Failed to fetch playlist information', async (spinner) => {
+    const playlistIds = options.playlistIds;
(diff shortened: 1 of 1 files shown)


prog893:
@coderabbitai resume

prog893:
@coderabbitai resume

PR #73, opened 2026-06-06
fix: validate --start-date / --end-date format across 5 commands

### LINKED ISSUE #61: Bug: get-video-analytics crashes silently when --start-date or --end-date is not YYYY-MM-DD
## Summary

Passing an invalid date string to `--start-date` or `--end-date` silently returns empty data with exit code 0 instead of showing a validation error. The user has no indication their date was malformed.

## Repro

```bash
staqan-yt get-video-analytics --video-id dQw4w9WgXcQ --start-date "not-a-date"
# Output: "Date Range: not-a-date to 2026-06-06 / No analytics data available"
# Exit: 0
```

## Root cause

`commands/get-video-analytics.ts:51-52` passes `options.startDate` / `options.endDate` directly to `chunkDateRange()` without format validation. The date string is accepted as-is and used as a display label while the API either ignores or rejects the malformed range.

## Expected behavior

```
✗ --start-date must be in YYYY-MM-DD format
```

## Fix

Validate before use:
```ts
const isValidDate = (d: string) => /^\d{4}-\d{2}-\d{2}$/.test(d) && !isNaN(new Date(d).getTime());
if (options.startDate && !isValidDate(options.startDate)) {
  error('--start-date must be YYYY-MM-DD'); process.exit(1);
}
```

Confirmed with runtime testing.

prog893:
## Summary

- Adds `parseDateOption(opt, flag)` helper to `lib/utils.ts` — throws if value is present but not `YYYY-MM-DD`, callers catch-and-die (same pattern as `parsePositiveInt`)
- Applies validation before `withSpinner` in all 5 affected commands: `get-video-analytics`, `get-channel-analytics`, `get-channel-search-terms`, `fetch-reports`, `get-report-data`
- When both flags are provided, also rejects reversed ranges (`--start-date > --end-date`) before any API call

## Test plan

- [x] `--start-date "not-a-date"` → `✗ --start-date must be in YYYY-MM-DD format`, exit 1
- [x] `--start-date 2026-01-01 --end-date 2025-01-01` → `✗ --start-date must be before --end-date`, exit 1
- [x] Invalid month (`2025-13-01`) rejected by `Date.parse` check
- [x] Valid date passes through to API call as before
- [x] All 5 commands tested manually

Closes #61

### DIFF
--- a/commands/fetch-reports.ts
+++ b/commands/fetch-reports.ts
@@ -1,6 +1,6 @@
 import { google } from 'googleapis';
 import { getAuthenticatedClient } from '../lib/auth';
-import { error, debug, success, warning, initCommand, withSpinner, formatTimestampWithTimezone } from '../lib/utils';
+import { error, debug, success, warning, initCommand, withSpinner, formatTimestampWithTimezone, parseDateOption } from '../lib/utils';
 import {
   isReportCached,
   saveReportToCache,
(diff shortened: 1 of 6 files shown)


prog893:
Superseded by #75, which uses deterministic calendar-math validation, consistent throw-catch-die pattern, and correct inclusive range check.

PR #74, opened 2026-06-06
fix: rename --tags to --replace and validate mutual exclusivity with --add/--remove

### LINKED ISSUE #64: Bug: update-video-tags silently ignores --add/--remove when --tags is also provided
## Summary

Providing `--tags` together with `--add` or `--remove` does not produce an error — the conflicting flags are silently ignored, leading to unexpected results.

## Repro

```bash
staqan-yt update-video-tags --video-id ID --tags "foo,bar" --add "baz" --dry-run
# --add "baz" is silently ignored; only --tags applies
```
Exit: 0 (no conflict detected)

## Root cause

`commands/update-video-tags.ts` has no guard that errors when `--tags` is combined with `--add` or `--remove`. These are mutually exclusive modes (replace vs. append/remove) but the CLI accepts both silently.

## Fix

Add early validation:
```ts
if (options.tags && (options.add || options.remove)) {
  error('Cannot combine --tags with --add or --remove. Use one mode at a time.');
  process.exit(1);
}
```

prog893:
## Summary

**BREAKING CHANGE (part of 2.1.0 milestone)**

- Renames `--tags` to `--replace` for clarity — rewrite mode is now explicit at the flag name level
- Adds validation: `--replace` cannot be combined with `--add` or `--remove` (previously silent ignore)
- Improves permission error message — explains "you don't own this video" vs "invalid tag characters" instead of misleading "invalid keywords"

## Root cause

Original `--tags` flag had silent failure when combined with `--add`/`--remove` because the code branch never reached them:

```typescript
if (options.tags) {
  newTags = options.tags.split(',')...  // --add/--remove never reached
} else {
  if (options.add) { ... }
  if (options.remove) { ... }
}
```

## Breaking change

**Before:**
```bash
staqan-yt update-video-tags --video-id ID --tags "foo,bar"
```

**After (2.1.0):**
```bash
staqan-yt update-video-tags --video-id ID --replace "foo,bar"
```

**Rationale:** `--replace` makes the mode (rewrite all vs incremental add/remove) immediately clear from the flag name. This aligns with the project's goal of consistent, discoverable CLI ergonomics.

## Test plan

- [x] `--replace "foo,bar"` → rewrite mode works
- [x] `--replace "foo" --add "bar"` → `✗ --replace cannot be combined with --add or --remove`, exit 1
- [x] `--add "tag"` alone → incremental mode works
- [x] `--add "x" --remove "y"` together → works (add + remove in one pass)
- [x] Old `--tags` flag → `Error: unknown option '--tags'` (breaking change confirmed)
- [x] Unowned video write → clear error explaining both possible causes

Closes #64

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the replace flag so it accepts its own arguments and no longer conflicts with other tag options.
  * Enforced that replace cannot be used with add or remove.

* **Improvements**
  * Better, multi-line error messages and a tags input tip for invalid keyword errors.

* **Documentation**
  * Updated docs, examples, and shell completions to reflect the new replace flag.

* **Breaking Changes**
  * CLI now uses --replace (replacing --tags) and rejects combined replace/add/remove usage.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -1,5 +1,23 @@
 # Breaking Changes
 
+## v2.1.0
+
+**`--tags` → `--replace`** in `update-video-tags`:
+
+| Before | After |
+|---|---|
+| `--tags "foo,bar"` | `--replace "foo,bar"` |
+
+**Why:** `--tags` was ambiguous — `--replace` makes it clear this replaces all tags.
+
+**Also:** Mixing modes now errors (was silently ignored before):
+```bash
+# ✗ --replace cannot be combined with --add or --remove
+staqan-yt update-video-tags --video-id ID --replace "foo" --add "bar"
+```
+
+---
+
 ## v2.0.0: Channel-isolated cache and data storage
 
 **Directory structure change** — all per-channel data now lives under a channel ID namespace:
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -248,7 +248,7 @@ staqan-yt get-video-analytics --video-id dQw4w9WgXcQ
 
 # Get/update tags
 staqan-yt get-video-tags --video-id dQw4w9WgXcQ
-staqan-yt update-video-tags --video-id dQw4w9WgXcQ --tags "tag1,tag2"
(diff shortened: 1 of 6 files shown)


prog893:
@coderabbitai resume

PR #79, opened 2026-06-18
fix: get-caption outputs Blob instead of caption content (#53)

### LINKED ISSUE #53: Bug: get-caption outputs `Blob (N KB)` instead of caption content
## Problem

`staqan-yt get-caption --caption-id <id>` prints `Blob (18.1 KB)` instead of the actual caption text.

## Root Cause

In `lib/youtube.ts:912`, `downloadCaption()`:

```typescript
const response = await youtube.captions.download({ id: captionId });
const content = response.data as string;
return content;
```

`response.data` from the Google API YouTube client is a `Blob` object (when `responseType` is not explicitly set), not a string. Casting it with `as string` just calls `.toString()` on the Blob, which produces `"Blob (18.1 KB)"`.

## Fix

Either:
1. Set `responseType: "text"` on the API request so `response.data` is already a string
2. Or read the Blob: `const content = await (response.data as Blob).text()`

## Reproduction

```bash
staqan-yt list-captions --video-id moYDTCX0GO8
# Copy any en caption ID
staqan-yt get-caption --caption-id <id> --format srt
# Output: Blob (18.1 KB)
```

## Environment
- staqan-yt 2.0.7
- Node v22.20.0
- macOS arm64

prog893:
## Summary

- `captions.download()` was returning a `Blob` object when `responseType` was not set
- Casting `response.data as string` called `.toString()` on the Blob, producing `"Blob (18.1 KB)"`
- Fix: pass `{ responseType: 'text' }` as the second argument so `response.data` is already a string

Fixes #53

## Test plan

- [ ] `staqan-yt list-captions --video-id <id>` — copy a caption ID
- [ ] `staqan-yt get-caption --caption-id <id>` — should output caption text, not `Blob (N KB)`
- [ ] `staqan-yt get-caption --caption-id <id> --format srt` — same check for SRT format

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved caption retrieval to ensure proper text format handling from the API.

### DIFF
--- a/lib/youtube.ts
+++ b/lib/youtube.ts
@@ -917,7 +917,7 @@ async function downloadCaption(captionId: string, format: CaptionFormat = 'json'
   // The download endpoint returns the caption content
   const response = await youtube.captions.download({
     id: captionId,
-  });
+  }, { responseType: 'text' });
 
   // The response is the actual caption content
   const content = response.data as string;


prog893:
@coderabbitai review

prog893:
@coderabbitai review

PR #89, opened 2026-06-30
fix: hoist jobs.list() and add RPM retry to fetch-reports

### LINKED ISSUE #84: fetch-reports: jobs.list() called per report type + no RPM retry causes quota exhaustion
## Problem

`fetch-reports` consistently hits YouTube Reporting API's **per-minute** quota (`Free requests per minute`) when there's any backlog to process. Two root causes:

### 1. `jobs.list()` called inside the report-type loop (20× unnecessary calls)

In `commands/fetch-reports.ts` line 185, `youtubeReporting.jobs.list()` is called **once per report type** in the loop:

```ts
for (const reportType of reportTypes) {
  // ...
  const jobsResponse = await youtubeReporting.jobs.list({ ... }); // called 20 times!
  const jobs = jobsResponse.data.jobs || [];
  const matchingJob = jobs.find(job => job.reportTypeId === reportTypeId);
```

This fetches the full jobs list (same 20 jobs every time) for each of the 20 report types = **20 identical API calls** instead of 1. Moving this call outside the loop would cut API usage by ~95% for the listing phase.

### 2. No error differentiation or retry on rate limit

The `downloadReport` function (line 324) catches all errors the same way with no retry. When YouTube returns 429, the error message text **does** distinguish RPM vs daily quota:

- **RPM (retriable):** `Quota exceeded for quota metric 'Free requests' and limit 'Free requests per minute'`
- **Daily (bail out):** `Quota exceeded for quota metric 'Free requests' and limit 'Free requests per day'`

Currently both are treated as hard failures with no retry. An RPM 429 just needs a ~60s backoff and retry; a daily quota 429 should abort immediately.

## Evidence

From cron run history: 8 of the last 14 runs failed with `Quota exceeded ... 'Free requests per minute'` — consistently the RPM limit, not daily exhaustion.

## Suggested fix

1. **Hoist `jobs.list()` outside the `for (const reportType of reportTypes)` loop** — call once, build a lookup map by `reportTypeId`.

2. **In `downloadReport` (and the API call error handler), parse the 429 error message:**
   - If `per minute` → sleep 60s and retry (up to N times)
   - If `per day` → throw immediately with a clear message

## Impact

Daily cron (`yt-fetch-reports-daily`) fires at 6AM JST and fails on any day with new reports to download due to #1 alone.

prog893:
Fixes #84 — fetch-reports daily cron fails 8/14 runs from quota exhaustion.

## Root causes

1. **jobs.list() called inside the report-type loop** (~20× per run instead of 1×). Each iteration re-fetched the identical jobs list.
2. **No differentiation between RPM and daily quota** — RPM 429 just failed immediately, and any retry would have used the wrong backoff (1s/2s/4s exponential from `retryWithBackoff`, not the ~60s RPM window reset).

## Fix

**commands/fetch-reports.ts:**
- Hoist `jobs.list()` out of the loop, index results into a `Map<reportTypeId, Job>`. Cuts listing-phase API usage by ~95% for the default 20-type run.
- Wrap every googleapis call (`reportTypes.list`, `jobs.list`, `jobs.create`, `jobs.reports.list`) with `withRateLimitRetry()`.
- Refactor `downloadReport` into `downloadOnce` + retry loop. Retries on HTTP 429 from the download host with exponential backoff (5s→10s→20s→40s→80s, capped 90s), honoring `Retry-After` header. Also retries on transient network errors.

**lib/utils.ts (new helpers, reusable):**
- `isRateLimitError(err)` → `'rpm' | 'daily' | null`. Parses both `err.message` and nested `err.response.data.error.errors[]` for the YouTube "Free requests per minute" vs "Free requests per day" messages.
- `withRateLimitRetry(fn, opts)` → exponential backoff on RPM, immediate throw on daily, pass-through on everything else. Defaults: 5 retries, 5s base / 90s max delay.

## Verified

```
[DEBUG] Indexed 20 existing job(s) by reportTypeId
[DEBUG] Found existing job: 3ed3845f-76fa-4bfe-9b1d-c286deaffa84
...
✔ channel_basic_a3: 0 downloaded, 74 cached, 0 errors
✓ Fetch complete!
```

- `npm run type-check` — pass
- `npm run lint` — pass (no new warnings)
- `npm run build` — pass
- End-to-end against real Reporting API with `--types` filter — pass

## Summary by CodeRabbit

* **New Features**
  * Added smarter retry handling for report downloads and API calls when rate limits or temporary network issues occur.
  * Improved report fetching to reuse existing job data within a run, reducing unnecessary repeated requests.

* **Bug Fixes**
  * Better recovery from HTTP 429 responses by honoring retry timing and cleaning up incomplete downloads.
  * Added support for transient connection failures during downloads, helping report retrieval complete more reliably.

* **Documentation**
  * Updated reporting quota and rate-limit guidance to reflect the new retry behavior and request usage.

### DIFF
--- a/commands/fetch-reports.ts
+++ b/commands/fetch-reports.ts
@@ -15,6 +15,7 @@ import { acquireLock, getLockPath } from '../lib/lock';
 import https from 'https';
 import { createWriteStream, unlinkSync } from 'fs';
 import { promises as fs } from 'fs';
+import { pipeline } from 'stream/promises';
 import { unlink } from 'fs/promises';
 import path from 'path';
 import chalk from 'chalk';
@@ -28,6 +29,18 @@ interface FetchReportsOptions {
   force?: boolean;
   verify?: boolean;
   verbose?: boolean;
+}
+
+/**
+ * Build a filesystem-safe temp path for a report download.
+ *
+ * `report.id` comes from an external API response and is interpolated into a
+ * temp filename. Strip anything outside [A-Za-z0-9._-] so a malicious or
+ * malformed ID (e.g. "../foo" or "id/with/slashes") can't escape /tmp.
+ */
+function safeTmpReportPath(reportId: string | null | undefined): string {
+  const safeId = String(reportId ?? 'report').replace(/[^A-Za-z0-9._-]/g, '_');
+  return path.join('/tmp', `${safeId}.csv`);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


prog893:
@coderabbitai review

PR #85, opened 2026-06-29
fix: reject invalid flags instead of silently falling back

prog893:
## Summary

All 5 remaining silent-validation bugs from the June audit, fixed in one pass:

- **#68** `config get <bad-key>` — was printing `not set` for typos; now errors with the valid-key list (same as `config set` already did)
- **#66** `list-comments --sort <garbage>` — silently mapped anything that wasn't `new` to `relevance`; now errors immediately  
- **#70** `get-channel-analytics --report X --dimensions Y` — `--dimensions`/`--metrics` were silently ignored when `--report` was present; now exits with a clear conflict message
- **#71** `fetch-reports --types ""` — empty string is falsy so the filter branch was skipped and **all** report types were downloaded; now errors on empty/whitespace-only `--types`
- **#65** `get-thumbnail --quality` — confirmed this was a false positive; `download-thumbnail` (added in #83) already validates quality correctly; `get-thumbnail` has no `--quality` flag

## Test plan

- [x] `staqan-yt config get bad.key` → error listing valid keys ✅
- [x] `staqan-yt list-comments --video-id <id> --sort garbage` → error listing valid values ✅
- [x] `staqan-yt get-channel-analytics --channel @x --report demographics --dimensions "country"` → conflict error ✅
- [x] `staqan-yt fetch-reports --types ""` → empty-types error (no API calls made) ✅
- [x] All valid usages of the above commands still work as expected ✅

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #90, opened 2026-07-01
fix(#88): filter --content-type client-side by video duration

### LINKED ISSUE #88: Bug: get-channel-search-terms --content-type video/shorts sends invalid filter to Analytics API
## Summary

\`staqan-yt get-channel-search-terms --content-type video\` and \`--content-type shorts\` both pass client-side validation (after #87) but the YouTube Analytics API rejects the request:

```
✗ Invalid value (LONG_FORM_CONTENT) given in field parameters.filters.
```

Exit code is 0 and the spinner still shows "Search terms data retrieved" because the error surfaces as an API exception inside the spinner, but **no data is returned**.

## Repro

```bash
staqan-yt get-channel-search-terms --content-type video
# API error: Invalid value (LONG_FORM_CONTENT) given in field parameters.filters.
# Exit: 0

staqan-yt get-channel-search-terms --content-type shorts
# API error: Invalid value (SHORT_FORM_CONTENT) given in field parameters.filters.
# Exit: 0
```

Only `--content-type all` (or no flag) actually works.

## Root cause

`commands/get-channel-search-terms.ts` builds the filter as:

```ts
const CONTENT_TYPE_FILTERS: Record<string, string> = {
  video: 'creatorContentType==LONG_FORM_CONTENT',
  shorts: 'creatorContentType==SHORT_FORM_CONTENT',
};
```

But `creatorContentType` is **not a valid dimension for the `insightTrafficSourceDetail` + `insightTrafficSourceType==YT_SEARCH` report**. The Analytics API rejects it with the 400 above.

Possible fixes (need investigation):
- Find the correct filter syntax for filtering YT_SEARCH traffic by content type for this report — the dimension name or operator may differ
- Or: scope the filter to per-video (fetching only video / only shorts playlists before calling Analytics) — but that breaks the report shape
- Or: drop content-type filtering for this report and surface a "not supported" warning instead of silently sending a broken filter

## Related

- #67 was the silent-fallback half of this bug — fixed in PR #87
- This issue tracks the remaining broken-API half

Discovered while verifying #87 against the actual Analytics API.

prog893:
## Summary

`staqan-yt get-channel-search-terms --content-type video|shorts` was throwing a 400 (`Invalid value (...) given in field parameters.filters`) because the YouTube Analytics API does **not** accept `creatorContentType` as a filter for the `insightTrafficSourceDetail` + `YT_SEARCH` report.

The error wording ("Invalid value") is misleading — it's the **dimension**, not the value, that isn't supported for this report type. Even swapping to the correct enum values (`VIDEO_ON_DEMAND` / `SHORTS`) still returns the same error. The only way to scope search traffic by content type for this report is to **pre-filter the video list by duration client-side**.

## Changes

- Drop the dead `creatorContentType` API filter entirely
- When `--content-type` is `video` or `shorts`, fetch durations via `videos.list` (chunked at 50 IDs per call, the documented limit) and trim `videoIds`:
  - **shorts**: duration < 60s
  - **video**: duration >= 60s
- Reuse the existing `parseDuration` helper from `lib/utils.ts`
- Update content-type validation to use a hardcoded allowlist (since the old filter map is gone)
- Apply the same fix to the MCP wrapper (`commands/mcp.ts`)
- Preserve the existing user-facing labels ("Regular videos only" / "Shorts only") and output shape
- Add a guard for the edge case where the channel has zero matching videos after filtering

## Verified live

```
staqan-yt get-channel-search-terms --content-type video  --limit 3
  Videos covered: 17

staqan-yt get-channel-search-terms --content-type shorts --limit 3
  Videos covered: 27

staqan-yt get-channel-search-terms --content-type all    --limit 3
  Videos covered: 44
```

17 + 27 = 44 — the math balances perfectly. Previously all three filtered calls returned a 400 error from the Analytics API.

## Followups (out of scope)

- `creatorContentType` is listed as a valid filter in the docs for many other reports (channel-level basics, geography, etc.). If we ever add content-type filtering to those, the correct enum values are `LIVE_STREAM`, `SHORTS`, `STORY`, `VIDEO_ON_DEMAND` — never the placeholder names we had here.

Closes #88

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved channel search-term reporting for Shorts and long-form videos by filtering videos based on actual duration.
  * Results now use a more reliable content-type check and return an error when no videos match the selected type.
  * Search-term queries now apply the requested video scope consistently before generating analytics results.

### DIFF
--- a/commands/get-channel-search-terms.ts
+++ b/commands/get-channel-search-terms.ts
@@ -1,16 +1,10 @@
 import chalk from 'chalk';
 import { getAuthenticatedClient } from '../lib/auth';
 import { google } from 'googleapis';
-import { parseChannelHandle, error, parsePositiveInt, debug, formatNumber, convertToCSV, initCommand, withSpinner, toLocalYmd, validateDateOption, validateDateRange } from '../lib/utils';
+import { parseChannelHandle, error, parsePositiveInt, debug, formatNumber, convertToCSV, initCommand, withSpinner, toLocalYmd, validateDateOption, validateDateRange, parseDuration } from '../lib/utils';
 import { getOutputFormat, requireChannel } from '../lib/config';
 import { formatJson, formatTable, formatCsv } from '../lib/formatters';
 import { ChannelSearchTermsOptions } from '../types';
-
-// Content type filter values for YouTube Analytics API
-const CONTENT_TYPE_FILTERS: Record<string, string> = {
-  video: 'creatorContentType==LONG_FORM_CONTENT',
-  shorts: 'creatorContentType==SHORT_FORM_CONTENT',
-};
 
 // Metrics for insightTrafficSourceDetail with insightTrafficSourceType==YT_SEARCH.
 // videoThumbnailImpressions/CTR are only valid for discovery-type sources and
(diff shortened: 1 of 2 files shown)


prog893:
✅ Applied. Pushed `1f40309` — `commands/mcp.ts` now validates `contentType` against `['all', 'video', 'shorts']` and throws on unknown values, mirroring the CLI's allowlist check. The `any`-typed `args` was the same silent-fallback bug class the project tackled in #85/#86, so this is the right fix.

prog893:
@coderabbitai review

Issue #78, opened 2026-06-07
housekeeping: validatePrivacyFilter exits directly instead of throwing


prog893:
## Problem

\`validatePrivacyFilter\` in \`lib/utils.ts\` calls \`error()\` + \`process.exit(1)\` directly instead of throwing:

\`\`\`ts
function validatePrivacyFilter(privacy: string[] | undefined): void {
  ...
  error(\`Invalid privacy value(s): \${invalid.join(', ')}. Valid values: public, private, unlisted\`);
  process.exit(1);  // ← helper should not exit
}
\`\`\`

This is inconsistent with \`parsePositiveInt\`, \`validateDateOption\`, and \`validateDateRange\`, which all throw and let callers catch-and-die.

## Fix

Change to throw, update all callers to catch-and-die:

\`\`\`ts
function validatePrivacyFilter(privacy: string[] | undefined): void {
  ...
  throw new Error(\`Invalid privacy value(s): \${invalid.join(', ')}. Valid values: public, private, unlisted\`);
}
\`\`\`

**Priority:** low / housekeeping

prog893:
Fixed in #92 (merged 2026-07-01): `validatePrivacyFilter` now throws instead of calling `error()` + `process.exit(1)` directly; both call sites wrap it in try/catch. Closing — same multi-issue auto-link gap as #77.

Issue #77, opened 2026-06-07
design: authenticate() in lib/auth.ts exits directly instead of throwing


prog893:
## Problem

\`authenticate()\` in \`lib/auth.ts\` calls \`process.exit(1)\` directly when credentials are missing:

\`\`\`ts
async function authenticate(): Promise<OAuth2Client> {
  const credentials = await loadCredentials();
  if (!credentials) {
    error('Credentials file not found!');
    console.log('');
    info('To set up authentication:');
    // ... multi-line setup instructions ...
    process.exit(1);  // ← library function should not exit
  }
  ...
}
\`\`\`

A library function that returns \`OAuth2Client\` should not own the exit decision. Callers (commands) should decide whether and how to exit.

## Complication

The exit is coupled to multi-line UX messaging (setup instructions with URLs and step-by-step guidance). A naive throw would lose that messaging unless the caller reconstructs it or the error carries it.

## Options

- Throw a typed error (e.g. \`CredentialsMissingError\`) that carries the setup instructions as metadata, so callers can display them before exiting
- Or throw a plain error and move the setup instruction output to a shared helper that callers invoke in the catch block

**Priority:** low / design / housekeeping

prog893:
Fixed in #92 (merged 2026-07-01): `authenticate()` now throws `CredentialsMissingError` instead of exiting directly; `commands/auth.ts` catches it and prints the setup instructions. Closing — PR referenced `Closes #76, #77, #78` but GitHub only auto-linked #76.

PR #92, opened 2026-07-01
fix(#76,#77,#78): align helpers with throw-and-catch pattern

### LINKED ISSUE #76: housekeeping: parsePositiveInt hardcodes --limit in error message
## Problem

\`parsePositiveInt\` in \`lib/utils.ts\` hardcodes \`--limit\` in its thrown error regardless of which flag was actually passed:

\`\`\`ts
function parsePositiveInt(limitOpt: string | undefined, defaultValue: number): number {
  const n = limitOpt ? parseInt(limitOpt, 10) : defaultValue;
  if (isNaN(n) || n < 1) throw new Error('--limit must be a positive integer');
  return n;
}
\`\`\`

If reused for any flag other than \`--limit\`, the error message will be wrong.

## Fix

Add a \`flag\` parameter so the error names the actual failing flag (same pattern as \`validateDateOption\`):

\`\`\`ts
function parsePositiveInt(flag: string, limitOpt: string | undefined, defaultValue: number): number {
  const n = limitOpt ? parseInt(limitOpt, 10) : defaultValue;
  if (isNaN(n) || n < 1) throw new Error(\`\${flag} must be a positive integer\`);
  return n;
}
\`\`\`

All callers would pass their flag name as the first argument.

**Priority:** low / housekeeping

prog893:
Closes #76, #77, #78.

The project documents the helper pattern as 'validation helpers throw; callers catch-and-die' (lib/utils.ts validates it via parsePositiveInt, validateDateOption, validateDateRange) but three helpers broke that pattern by exiting the process directly. This made unit testing impossible and created silent inconsistencies.

### Changes

- **#76** `parsePositiveInt(flag, opt, default)` — now takes the flag name as a parameter so the thrown error names the actual failing flag (no longer hardcoded as `--limit`). All six call sites updated.
- **#78** `validatePrivacyFilter()` — now throws instead of calling `error()` + `process.exit(1)`. Both call sites (`list-videos`, `list-playlists`) wrap it in try/catch.
- **#77** `authenticate()` — now throws a typed `CredentialsMissingError` when the credentials file is missing. The multi-line setup instructions were factored into `printCredentialsSetupInstructions()` (also exported) so the caller's catch block in `commands/auth.ts` can print them in the exact same shape as before. The library function no longer owns the exit decision.

### Smoke tests (post-build)

- `list-videos --limit abc` → `✗ --limit must be a positive integer` ✓
- `search-videos --limit -5` → same clean error from a different command ✓
- `list-videos --privacy bogus` → `✗ Invalid privacy value(s): bogus. Valid values: public, private, unlisted` ✓
- `auth` (no creds path) — exercised via the build; runtime check requires no credentials file, which the dev environment has.

`npm run type-check`, `npm run lint` (0 errors; 11 pre-existing `any` warnings unrelated), `npm run build` all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication failures by showing clear setup instructions when credentials are missing.
  * Validation and parsing errors now report more specific command names, making limit and filter issues easier to understand.
  * Invalid privacy settings are now handled more consistently across listing commands, with clearer error messages and cleaner exits.

### DIFF
--- a/commands/auth.ts
+++ b/commands/auth.ts
@@ -1,4 +1,4 @@
-import { authenticate } from '../lib/auth';
+import { authenticate, CredentialsMissingError, printCredentialsSetupInstructions } from '../lib/auth';
 import { success, error } from '../lib/utils';
 
 async function authCommand(): Promise<void> {
@@ -12,6 +12,12 @@ async function authCommand(): Promise<void> {
   } catch (err) {
     console.log('');
     error(`Authentication failed: ${(err as Error).message}`);
+    // When the credentials file is missing, print the setup instructions so
+    // the user can fix it without a second guess about where to look.
+    if (err instanceof CredentialsMissingError) {
+      console.log('');
+      printCredentialsSetupInstructions();
+    }
     process.exit(1);
   }
 }
--- a/commands/fetch-reports.ts
+++ b/commands/fetch-reports.ts
@@ -1,6 +1,6 @@
 import { google } from 'googleapis';
 import { getAuthenticatedClient } from '../lib/auth';
-import { error, debug, success, warning, progress, initCommand, withSpinner, formatTimestampWithTimezone, validateDateOption, validateDateRange, withRateLimitRetry } from '../lib/utils';
+import { error, debug, success, warning, progress, initCommand, withSpinner, formatTimestampWithTimezone, validateDateOption, validateDateRange, withRateLimitRetry, runOrExit } from '../lib/utils';
 import {
   isReportCached,
   saveReportToCache,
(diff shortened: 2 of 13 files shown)


prog893:
@coderabbitai review --resume

prog893:
@coderabbitai resume

prog893:
@coderabbitai review

prog893:
Resolved in commit 92676f3 on this branch (pushed a few minutes before this review). The `runOrExit` helper was added to lib/utils.ts and all 24 inline validator call sites were converted:

- `lib/utils.ts`: new `runOrExit<T>(fn: () => T): T` helper (centralized try/catch → error/exit)
- 10 command files converted: list-videos, list-playlists, list-comments, search-videos, get-search-terms, get-channel-search-terms, get-video-analytics, get-channel-analytics, get-report-data, fetch-reports

The diff shown in this thread is the OLD inline pattern, which no longer exists in those files. The two spinner-context sites (get-channel-search-terms.ts:198, get-video-analytics.ts:58) also had their redundant manual `try/catch { spinner.stop(); error; exit }` removed — validateDateRange now throws straight to withSpinner's own catch-and-die.

`npm run type-check`, `npm run lint`, `npm run build` all pass. Smoke-tested all validator paths (--limit, --privacy, --start-date format, --start-date > --end-date) and the happy path (valid --privacy) — all behave correctly.

Could you re-review and resolve this CHANGES_REQUESTED, or confirm there are still blocking findings I missed? @coderabbitai resume

Issue #110, opened 2026-07-02
Bug: process.exit() in command error paths kills MCP server


prog893:
## Problem

Commands call `process.exit(1)` on errors. This is fine for CLI usage but makes the MCP server pattern brittle — `process.exit()` kills the entire MCP server process entirely when a command hits an error path.

## Reproduction

1. Start MCP server: `staqan-yt mcp`
2. Call any tool that triggers a command error (e.g., get-video with invalid ID)
3. MCP server process exits entirely
4. All subsequent MCP tool calls fail

## Proposed Solution

1. **Refactor commands to throw errors instead of calling `process.exit()`**:
   - Throw typed errors (`YouTubeAPIError`, `ValidationError`, etc.)
   - CLI layer catches and converts to exit code 1
   - MCP layer catches and converts to error response
2. **Add tests** for error handling in both CLI and MCP contexts
3. **Audit all `process.exit()` calls** and replace with throws where appropriate

---

## Inventory of every `process.exit()` site on `main`

Counted via `grep -rn "process\.exit" commands/ lib/ bin/`.

**Totals:** 95 sites
- Out of scope (intentional / different protocol): **12**
- Helper layer — exit-decision layer of throw-and-catch contract: **3**
- Helpers already refactored to throw by **PR #92**: **3** ✅
- Command-level direct exits — candidates for refactor: **~77**

### Out of scope — intentional exits

These are legitimate exit points that should NOT change.

- [x] `bin/staqan-yt.ts:58` — top-level help/version handler exit 0
- [x] `bin/staqan-yt.ts:71` — top-level completion-script stub exit 0
- [x] `bin/staqan-yt.ts:111` — top-level `help <cmd>` exit 0
- [x] `bin/staqan-yt.ts:125` — `showError()` helper exit (top-level CLI error formatter)
- [x] `bin/staqan-yt.ts:146` — `commander.helpDisplayed`/`commander.version` exit 0
- [x] `bin/staqan-yt.ts:530` — `printOutputSummary` exit 0
- [x] `bin/staqan-yt.ts:538` — same handler exit 0
- [x] `bin/staqan-yt.ts:549` — same handler exit 0
- [x] `bin/staqan-yt.ts:567` — same handler exit 0
- [x] `commands/complete.ts:64` — completion protocol exit 0 (different protocol; not an MCP path)
- [x] `commands/complete.ts:77` — completion protocol exit 0
- [x] `commands/complete.ts:147` — completion protocol exit 0

[...]

PR #121, opened 2026-07-02
refactor(#110): throw instead of exit in runOrExit/withSpinner

prog893:
First step in the #110 refactor. The two helper layers `runOrExit` (sync) and `withSpinner` (async) used to call `process.exit(1)` directly on caught errors. That kills the entire process — including any long-running caller such as the MCP server — when a tool hits an error path.

## Changes

- **`lib/utils.ts`** — `runOrExit` and both branches of `withSpinner` now:
  1. Print the error via `error()` (unchanged)
  2. Re-throw wrapped in a `__helperFormatted` marker so the CLI top-level catch can recognize "already printed"
  3. Preserve the original error on `.cause` for stack-trace debugging

  Adds two new exports: `isHelperFormattedError(err)` (consumed by the bin wrapper) and a small private `markFormatted` helper.

- **`bin/staqan-yt.ts`** — `withHelpWrapper` now wraps the action call in try/catch. On a thrown error it prints the message (if not already printed by a helper) and exits with code 1. New import: `isHelperFormattedError`.

## CLI behavior preservation

| Scenario | Before | After |
|---|---|---|
| `list-videos --limit abc` | `✗ --limit must be a positive integer`, exit 1 | Same, exit 1 |
| `list-videos --privacy bogus` | `✗ Invalid privacy value(s): bogus. Valid values: public, private, unlisted`, exit 1 | Same, exit 1 |
| `get-video-analytics --start-date=invalid` | `✗ --start-date must be in YYYY-MM-DD format (got: invalid)`, exit 1 | Same, exit 1 |
| `--version` | `2.0.11`, exit 0 | Same, exit 0 |
| `help` | Usage text, exit 0 | Same, exit 0 |

## Diff

```
 bin/staqan-yt.ts | 20 ++++++++++++++++---
 lib/utils.ts     | 59 +++++++++++++++++++++++++++++++++++++++++++++-----------
 2 files changed, 65 insertions(+), 14 deletions(-)
```

## Verification

- `bun run type-check` ✓
- `bun run lint` ✓ (0 errors; 11 pre-existing `any` warnings, unchanged)
- `bun run build` ✓
- Smoke-tested all the above scenarios — exit codes and error messages match the previous CLI behavior exactly, no double-printing.

## What this enables

Subsequent #110 PRs can convert the ~77 remaining direct `process.exit(1)` call sites in command code to `throw new Error(...)` — now safe because the helpers no longer rely on the caller to exit, and the top-level catch in `withHelpWrapper` will translate thrown errors to the same exit-1 behavior.

Closes part of #110 (helper layer only — direct exits in command code remain).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command error handling so failures are reported more consistently and the app exits cleanly with the correct status code.
  * Prevented duplicate error messages in help-related flows and spinner-driven operations.
  * Made validation and API errors surface clearly without causing unexpected process behavior.

### DIFF
--- a/bin/staqan-yt.ts
+++ b/bin/staqan-yt.ts
@@ -385,7 +385,7 @@ program
 program
   .command('mcp')
   .description('Start MCP server for AI assistant integration')
-  .action(mcpCommand);
+  .action(withHelpWrapper('mcp', mcpCommand));
 
 // Playlist commands
 // Get single playlist command (singular)
--- a/bin/staqan-yt.ts
+++ b/bin/staqan-yt.ts
@@ -4,7 +4,7 @@ import { program, Command } from 'commander';
 import chalk from 'chalk';
 import * as path from 'path';
 import { GroupedHelp } from '../lib/customHelp';
-import { setQuiet, setVerbose } from '../lib/utils';
+import { setQuiet, setVerbose, error, isHelperFormattedError } from '../lib/utils';
 import authCommand = require('../commands/auth');
 import listVideosCommand = require('../commands/list-videos');
 import getVideoCommand = require('../commands/get-video');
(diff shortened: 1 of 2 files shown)


prog893:
@coderabbitai resume

prog893:
@coderabbitai review

prog893:
@coderabbitai review

Issue #124, opened 2026-07-08
--output flag is dead on 6 commands; get-channel-analytics bypasses getOutputFormat


prog893:
## Problem

Two flavors of the same inconsistency, found by cross-checking `bin/staqan-yt.ts` registrations against command implementations:

1. **Dead flag**: `auth`, `config`, `update-video`, `update-video-tags`, `put-video-localization`, and `update-video-localization` all advertise `--output <format>` in help, but their implementations never read `options.output` at all. Users who pass `--output json` get no error and no JSON.

2. **Config bypass**: `get-channel-analytics` uses `options.output` directly (line ~187) instead of `await getOutputFormat(options.output)`. Consequences:
   - `config set default.output csv` is silently ignored by this command
   - an invalid value like `--output yaml` silently falls back to pretty — exactly the "reject invalid flags" class fixed elsewhere in #85

## Fix

- For the update/put commands: either implement `--output` (they return API results that can be formatted) or remove the flag from their registrations — pick one and be consistent.
- `get-channel-analytics`: switch to `getOutputFormat(options.output)` like the other 22 commands.

Found during 2026-07-08 codebase audit.

PR #133, opened 2026-07-08
fix(#124): validate --output centrally; make get-channel-analytics honor default.output

prog893:
Part of #124 (the config-bypass half; the dead-flag half stays open — see below).

## Value proposition

Two silent-fallback bugs, one fix point:

1. `get-channel-analytics` was the only `--output`-consuming command reading `options.output` raw instead of `await getOutputFormat(...)` — so `config set default.output csv` was silently ignored by this one command.
2. `getOutputFormat` itself never validated the flag, so `--output yaml` silently pretty-printed in **every** command. This is the same bug class #85 stamped out for `--sort`/`config get`; putting the check inside `getOutputFormat` fixes all 23 callers at once instead of per-command validation.

The format is now resolved **before** the spinner starts, so an invalid value costs zero API quota.

## Changes

- `lib/config.ts`: `getOutputFormat` rejects values outside `VALID_OUTPUT_FORMATS` with the same message `config set default.output` already uses.
- `commands/get-channel-analytics.ts`: resolve `outputFormat` via `getOutputFormat` at the top of the command; use it in the output switch.

## Verification (live)

- `get-channel-analytics --report geography --output yaml` → `✗ Invalid output format: yaml. Must be one of: json, table, text, pretty, csv`, exit 1, **no API call made**
- `config set default.output csv` then `get-channel-analytics --report geography` (no flag) → emits CSV (`country,views,estimatedMinutesWatched` / `JP,13,81`) — previously pretty-printed, ignoring config. Config reverted to `pretty` after the test.
- `bun run type-check` ✓ · `bun run lint` ✓ · `bun run build` ✓

## Remaining scope of #124

Six commands (`auth`, `config`, `update-video`, `update-video-tags`, `put-video-localization`, `update-video-localization`) advertise `--output` but never read it. Removing the flag would break scripts that pass it (Commander errors on unknown options); implementing it is real feature work per command. Left #124 open for that decision.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved output format handling for channel analytics.
  - Invalid output formats are now rejected with a clear error before analytics data is retrieved.
  - Supported formats and configured defaults are applied consistently across JSON, CSV, table, text, and pretty output.

### DIFF
--- a/commands/get-channel-analytics.ts
+++ b/commands/get-channel-analytics.ts
@@ -2,7 +2,7 @@ import chalk from 'chalk';
 import { getAuthenticatedClient } from '../lib/auth';
 import { google } from 'googleapis';
 import { parseChannelHandle, error, debug, formatNumber, initCommand, withSpinner, createSpinner, toLocalYmd, validateDateOption, validateDateRange, runOrExit } from '../lib/utils';
-import { requireChannel } from '../lib/config';
+import { requireChannel, getOutputFormat } from '../lib/config';
 import { formatJson, formatTable, formatCsv } from '../lib/formatters';
 import { ChannelAnalyticsOptions } from '../types';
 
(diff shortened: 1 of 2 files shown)


prog893:
@coderabbitai review

PR #135, opened 2026-07-08
fix(#127): single version source shared by CLI and MCP server

### LINKED ISSUE #127: Version-string drift: MCP server reports 1.3.0, CLAUDE.md footer says 2.0.7
## Problem

`scripts/version.ts` syncs the version to `bin/staqan-yt.ts` and the Homebrew formula on `bun pm version`, but two other version strings drift:

- `commands/mcp.ts` (~line 1418): MCP server registers as version `'1.3.0'` — clients see a version unrelated to the CLI release
- `CLAUDE.md` footer: "Current version: **2.0.7**" — actual is 2.0.11

## Fix

Either add both to `scripts/version.ts`'s sync list, or remove the hardcoded strings (CLAUDE.md already says "see package.json for source of truth" — the number adds nothing; mcp.ts can import the same resolved version the bin uses).

prog893:
Closes #127.

## Value proposition

Three version strings, three answers: the CLI said 2.0.11, the MCP server introduced itself to clients as **1.3.0** (hardcoded since it was written), and CLAUDE.md's footer said **2.0.7**. Any debugging session that cross-references MCP client logs against releases starts from a lie. One `getVersion()` makes drift structurally impossible rather than a thing to remember.

## Changes

- **`lib/version.ts` (new)**: `getVersion()` — package.json at runtime, `FALLBACK_VERSION` for compiled binaries. The fallback line carries a `// synced by scripts/version.ts` marker.
- **`bin/staqan-yt.ts`**: replaces its inline version block with `getVersion()` (also drops the now-unused `path` import).
- **`commands/mcp.ts`**: MCP server registers with `getVersion()` instead of `'1.3.0'`.
- **`scripts/version.ts`**: syncs `lib/version.ts` instead of `bin/staqan-yt.ts`, and now **throws if the marker line is missing** — previously a silent no-op regex miss would ship a stale fallback.
- **`CLAUDE.md`**: footer no longer duplicates the number; points at package.json (it already called package.json the source of truth while contradicting it two words earlier).

## Verification

- `staqan-yt --version` → `2.0.11` (built from this branch)
- Sync regex dry-run: replacing the marker line with a new version string matches exactly once; deleting it makes `scripts/version.ts` throw as intended.

## Summary by CodeRabbit

* **Improvements**
  * CLI and MCP server versions now stay synchronized automatically with the package version.
  * Added a fallback version for bundled or compiled CLI usage when package metadata is unavailable.
* **Documentation**
  * Updated version documentation to direct readers to `package.json` as the authoritative source.
* **Maintenance**
  * Improved the release process for updating and staging version information.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -453,4 +453,4 @@ async function getVideo() { /* API only */ }
 
 ## Version
 
-Current version: **2.0.7** (see `package.json` for source of truth)
+See `package.json` for the current version (single source of truth — the number is not duplicated here because it drifts).
--- a/bin/staqan-yt.ts
+++ b/bin/staqan-yt.ts
@@ -2,8 +2,8 @@
 
 import { program, Command } from 'commander';
 import chalk from 'chalk';
-import * as path from 'path';
 import { GroupedHelp } from '../lib/customHelp';
+import { getVersion } from '../lib/version';
 import { setQuiet, setVerbose, error, isHelperFormattedError } from '../lib/utils';
 import authCommand = require('../commands/auth');
 import listVideosCommand = require('../commands/list-videos');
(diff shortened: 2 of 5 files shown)


prog893:
@coderabbitai review

prog893:
Applied the finding: getVersion() now probes both layouts (`../../` for dist/lib, `../` for source lib/) and additionally verifies `name === 'staqan-yt-cli'` so a stray package.json in a parent directory can't hijack resolution. Verified both: built `--version` → 2.0.11, `bunx tsx bin/staqan-yt.ts --version` → 2.0.11.

Issue #125, opened 2026-07-08
Consolidate channel resolution: 4 independent implementations, only 1 uses the handle cache


prog893:
## Problem

Channel handle → ID resolution is implemented four separate times:

1. `lib/youtube.ts` `getChannelId()` — the only one with the disk handle-cache, but uses `search.list` (see #122)
2. `commands/get-channel-search-terms.ts` (~lines 57–96) — inline `forHandle` resolution
3. `commands/get-channel-analytics.ts` (~line 45+) — same inline pattern
4. `commands/mcp.ts` — two more inline copies (`youtube_get_channel_analytics`, `youtube_get_channel_search_terms`)

The three inline copies skip the handle cache entirely, so every invocation re-spends API quota on a lookup the cache already knows. They also each re-implement error handling slightly differently (some silently continue on empty responses — see #123).

## Fix

One `resolveChannel(handleOrId)` in `lib/youtube.ts` that: short-circuits UC… IDs, checks the disk cache, resolves via `channels.list({ forHandle })`, errors clearly on not-found, and caches the result. All four call sites use it.

This is also the prerequisite for shrinking `commands/mcp.ts` (1,459 lines, ~800 of which duplicate command logic — related to #102).

Found during 2026-07-08 codebase audit.

PR #136, opened 2026-07-08
fix(#123): channel URLs no longer misclassified as channel IDs

### LINKED ISSUE #123: parseChannelHandle misclassifies channel URLs (@-URLs, /c/, /user/ treated as channel IDs)
## Problem

`lib/utils.ts` `parseChannelHandle()` (lines ~35–58): for URL inputs, every matched pattern returns `{ type: 'id', value: match[1] }` — including:

- `youtube.com/@staqan` → `{ type: 'id', value: 'staqan' }` — this is a **handle** (and the `@` is stripped), not an ID
- `youtube.com/c/SomeName` and `youtube.com/user/SomeName` → legacy custom-URL/username segments, which are **not** channel IDs

Downstream, `get-channel-search-terms.ts` (lines ~87–96) calls `channels.list({ id: [value] })`, gets zero items, and **silently proceeds** with the bogus value as `channelId` — the Analytics query then fails opaquely or returns nothing. Same pattern in `get-channel-analytics.ts`.

## Fix

- `/@handle` URLs → `{ type: 'handle', value: '@' + match[1] }`
- `/c/` and `/user/` URLs → either resolve via `forUsername`/search, or reject with a clear "pass the @handle or UC… ID" error
- Empty `channels.list({ id })` responses should error out instead of continuing

Related: consolidating the four independent channel-resolution implementations would make this a single-point fix (separate issue).

prog893:
Closes #123.

## Value proposition

Passing a channel URL (the thing you most naturally copy from a browser) to `-c` silently broke: `youtube.com/@staqan` was classified as a channel *ID* (with the `@` stripped), `channels.list({id})` returned nothing, and both analytics commands **proceeded anyway** with the bogus value — surfacing later as an opaque Analytics error or empty data. Silent-wrong becomes either correct or loudly-wrong.

## Changes

- `lib/utils.ts` `parseChannelHandle`:
  - `youtube.com/@handle` URLs → `{ type: 'handle', value: '@handle' }` (resolves correctly via `forHandle`)
  - `/channel/UC…` URLs → ID, as before
  - `/c/Name` and `/user/Name` → **throw with guidance** ("Pass the channel's @handle or its UC… channel ID"). These legacy segments are neither handles nor IDs; every downstream lookup was guaranteed to fail, just less legibly.
- `commands/get-channel-search-terms.ts` + `commands/get-channel-analytics.ts`: an empty `channels.list` response now throws `Channel not found: <input>` instead of silently continuing. In `get-channel-analytics` this also removes a debug-only catch that swallowed lookup errors entirely (a channel the Analytics query can't use is not a cosmetic failure), and converts the handle-branch `process.exit(1)` to a throw (#110 pattern).

## Verification

Parse unit cases (compiled output): `@staqan` → handle · `https://www.youtube.com/@staqan` → handle · `/channel/UC…?sub=1` → ID (query string stripped) · raw `UC…` → ID · `/user/…` and `/c/…` → clear error.

Live on @staqan:
- `get-channel-analytics -c "https://www.youtube.com/@staqan" --report geography --output json` → resolves to STAQAN, returns data (**previously silently broken**)
- `get-channel-analytics -c "https://youtube.com/user/oldname" …` → legacy-URL error, exit 1, zero API quota spent

Related: #125 (consolidating the four channel-resolution implementations) would make this class of fix single-point in future.

## Summary by CodeRabbit

* **Bug Fixes**
  * Channel analytics and search-term commands now clearly report an error when the requested channel cannot be found.
  * Channel lookup failures no longer continue silently or produce confusing downstream errors.
  * YouTube handle and channel URL inputs are now classified more accurately.
  * Legacy `/c/` and `/user/` URL formats now return a clear unsupported-format error.

### DIFF
--- a/commands/get-channel-analytics.ts
+++ b/commands/get-channel-analytics.ts
@@ -54,40 +54,37 @@ async function getChannelAnalyticsCommand(options: ChannelAnalyticsOptions): Pro
     let channelTitle = '';
     let actualChannelId = parsedChannel.value;
 
-    try {
-      // Resolve channel handle to ID if needed
-      if (parsedChannel.type === 'handle') {
-        debug('Looking up channel by handle:', parsedChannel.value);
-        const channelResponse = await youtube.channels.list({
-          part: ['id', 'snippet'],
-          forHandle: parsedChannel.value.replace('@', ''),
-        });
+    // Resolve channel handle to ID if needed. A not-found channel fails the
+    // command here — previously the ID branch swallowed the empty response
+    // (and any lookup error) into a debug-only catch and proceeded, so the
+    // Analytics query failed later with an opaque message (#123).
+    if (parsedChannel.type === 'handle') {
+      debug('Looking up channel by handle:', parsedChannel.value);
+      const channelResponse = await youtube.channels.list({
+        part: ['id', 'snippet'],
+        forHandle: parsedChannel.value.replace('@', ''),
+      });
 
-        if (channelResponse.data.items && channelResponse.data.items.length > 0) {
-          actualChannelId = channelResponse.data.items[0].id!;
-          channelTitle = channelResponse.data.items[0].snippet?.title || '';
-        } else {
-          spinner.fail('Channel not found');
-          error(`Channel not found: ${channelId}`);
-          process.exit(1);
-        }
-      } else {
-        // Get channel by ID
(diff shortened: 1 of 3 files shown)


prog893:
@coderabbitai review

PR #134, opened 2026-07-08
refactor(#126): consolidate retry helpers — migrate to withRateLimitRetry, delete retryWithBackoff

prog893:
Closes #126.

## Value proposition

Two retry helpers doing the same job means every new API call site has to pick one, and the older one picks wrong: `retryWithBackoff` string-matched `'quota'/'429'`, kept retrying even when the **daily** quota was exhausted (wasting minutes and hiding the real problem), and ignored the server's `Retry-After`. `withRateLimitRetry` (built for #84) gets all three right. One helper, one behavior.

## Changes

- `commands/get-video-analytics.ts`, `commands/get-video-retention.ts`, `commands/mcp.ts` (video-analytics tool): migrate to `withRateLimitRetry` with per-call labels so backoff logs say what they're retrying.
- `lib/utils.ts`: delete `retryWithBackoff` (+ export). While in the file, three hygiene fixes flagged in #126:
  - The JSDoc block that described `validatePrivacyFilter` sat above `parsePositiveInt` (and still said helpers "exit" — they throw since #121). Each function now carries its own accurate doc.
  - `isRateLimitError`: dropped the dead duplicate condition (`'per minute'` OR-ed with `'per minute '`).
  - `withSpinner`: the quiet-mode silent-spinner object literal duplicated `createSpinner`'s — now one code path via `createSpinner`, −40 lines, identical output in both modes (double-print guard preserved).

**Deliberate behavior change**: quota errors that don't carry YouTube's per-minute wording are no longer blindly retried — RPM errors back off (honoring `Retry-After`), daily-quota errors abort immediately with a clear message, everything else surfaces at once instead of after 3 silent retries.

## Verification (live on @staqan)

- `get-video-analytics --video-id eeYl2dxv57g --output json` → valid payload (videoId/title/dateRange/columnHeaders/rows), exit 0
- `get-video-retention --video-id eeYl2dxv57g --output json` → 100 retention rows, exit 0
- `bun run type-check` ✓ · `bun run lint` ✓ (0 errors, 11 pre-existing warnings) · `bun run build` ✓

Note: GitHub Actions quota exhausted until end of month — no status checks; verified locally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved retry handling for video analytics and retention requests when rate limits occur.
  * Added clearer request identification during retry operations to improve error reporting.
  * Improved command progress indicators and failure behavior, including quieter output in quiet mode.
[...]

PR #138, opened 2026-07-11
fix(#137): rate-limit retry parity for get-report-data

### LINKED ISSUE #137: get-report-data: wrap Reporting API calls in withRateLimitRetry (parity with fetch-reports)
## Problem

`fetch-reports` wraps every Reporting API call (`reportTypes.list`, `jobs.list`, `jobs.create`, `jobs.reports.list`) in `withRateLimitRetry` since #89. `get-report-data` makes the same calls **unwrapped** (`jobs.list` ~line 69, `jobs.create` ~line 85, `jobs.reports.list` pagination loop ~line 115) — so the command that the daily CTR workflow actually reads data through is the one that dies on the first RPM 429 instead of backing off.

## Fix

Wrap the three call sites with `withRateLimitRetry(fn, { label })`, same pattern as `fetch-reports`. After #126 (PR #134) merges, this is mechanical.

Found during 2026-07-08 audit; filed separately to keep PRs #132/#134 focused.

prog893:
Closes #137.

## Value proposition

`fetch-reports` got proper RPM backoff in #89, but `get-report-data` — the command the daily CTR workflow actually reads through — kept making the same three Reporting API calls bare. One 429 at the wrong moment killed the read. Now both commands share identical retry semantics: exponential backoff with `Retry-After` honored on per-minute limits, immediate abort with a clear message on daily exhaustion.

## Changes

`commands/get-report-data.ts` only: `jobs.list`, `jobs.create`, and the `jobs.reports.list` pagination loop wrapped in `withRateLimitRetry` with per-call labels — the exact pattern `fetch-reports` uses. Mechanical now that #134 (retry consolidation) is merged.

## Verification

- Live: `get-report-data --type channel_reach_basic_a1 --start-date 2026-06-28 --end-date 2026-06-30 --output json` → valid rows, exit 0

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of temporary rate limits when retrieving YouTube reporting data.
  * Added automatic retries for discovering reporting jobs, creating missing jobs, and loading report pages.
  * Reduced failures caused by API backoff requirements during report retrieval.

### DIFF
--- a/commands/get-report-data.ts
+++ b/commands/get-report-data.ts
@@ -1,6 +1,6 @@
 import { google } from 'googleapis';
 import { getAuthenticatedClient } from '../lib/auth';
-import { warning, debug, initCommand, withSpinner, formatTimestampWithTimezone, validateDateOption, validateDateRange, runOrExit } from '../lib/utils';
+import { warning, debug, initCommand, withSpinner, formatTimestampWithTimezone, validateDateOption, validateDateRange, runOrExit, withRateLimitRetry } from '../lib/utils';
 import { getOutputFormat, getConfigValue } from '../lib/config';
 import { formatJson, formatTable, formatCsv, formatText } from '../lib/formatters';
 import {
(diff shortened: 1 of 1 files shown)


prog893:
@coderabbitai review

Issue #111, opened 2026-07-02
put-video-localization: add --video-ids flag for batch localization


prog893:
## Feature

Currently `put-video-localization` can only localize one video at a time. For translating a batch of 10 videos, that's 10 separate API calls. STAQAN is preparing to localize 3 JCBG episodes + 2 on-location docs into EN/JA/RU. That's 15 calls minimum.

## Current State

`put-video-localization --video-id ID --lang ja --title "..." --description "..."` - one video at a time.

## Proposed Solution

Add `--video-ids` flag (comma-separated or repeated) for batch operations:

```bash
# Batch localize 3 videos with same title/description
staqan-yt put-video-localization \
  --video-ids fn_vD4wB6lM,moYDTCX0GO8,2yQ7uxd7MVo \
  --lang ja \
  --title "タイトル" \
  --description "説明"
```

```bash
# Or read titles/descriptions from JSON file
staqan-yt put-video-localization --from-json ./localizations.json
```

Where `localizations.json` is:
```json
{
  "videos": [
    {"videoId": "fn_vD4wB6lM", "lang": "ja", "title": "...", "description": "..."},
    {"videoId": "moYDTCX0GO8", "lang": "ja", "title": "...", "description": "..."}
  ]
}
```

## Benefits

- Reduces 15 API calls to 1 batch call
- Easier to maintain consistency across episodes
- JSON file enables version control of localization data
- Dry-run support already exists for the single-video command

## References

- Documented in `staqan/research/staqan-yt-cli-architecture.md` (file 61) as "Actionable Insight #3: Missing batch localization"
- Related closed: #62 "put-video-localization and update-video-localization display 'null' in success message" (fixed display, not batch support)

Issue #102, opened 2026-07-02
MCP: replace console.log monkey-patching with structured return values


prog893:
## Problem

The MCP server in `commands/mcp.ts` captures command output by monkey-patching `console.log` and `console.error`. This is brittle:
- Commands that write to `process.stdout.write()` directly bypass capture
- Race conditions possible if multiple MCP tools run concurrently
- Hidden control flow makes debugging hard
- Couples MCP tightly to console-based output

## Current State

`commands/mcp.ts` (1291 lines) wraps each command by capturing console output. If a command writes directly to stdout/stderr (e.g., for progress bars, raw data), the MCP tool gets incomplete output.

## Proposed Solution

Refactor commands to return structured data (typed objects) instead of writing to console. MCP server serializes the return value. CLI layer adds a thin wrapper that formats the return value with `chalk`/`ora` for human display.

Benefits:
- No more console monkey-patching
- Concurrent MCP tool calls safe
- Easier to test (commands become pure functions)
- Clean separation of data layer and presentation layer

## Migration Path

1. Create typed return interfaces for each command
2. Migrate commands one at a time (start with read-only commands like get-video, list-videos)
3. Update MCP server to consume structured returns
4. Update CLI layer to format returns
5. Keep console output for backward compatibility during migration

## References

- Documented in `staqan/research/staqan-yt-cli-architecture.md` (file 61) as "Potential Issue #2: MCP console.log capture is fragile"

prog893:
**Progress + phase plan (2026-07-13):**

- ✅ **Phase 1 merged-pending-review as PR #146**: `lib/analytics.ts` + `fetchVideoAnalytics()` shared by the CLI command and the MCP tool. Pattern proven: lib returns typed data with an optional `onProgress` hook; command layers spinner/formats; MCP serializes. The MCP tool gained `dimensions` + `title` purely by deleting its copy — exactly the drift this issue predicted.

**Phase 2 (traffic-sources, search-terms, retention) — heads-up for the implementer:** these aren't mechanical extractions; the CLI and MCP copies have *diverged behavior* that consolidation must resolve deliberately:

1. **Date ranges**: the MCP copies query `2000-01-01 → today` in a single request; the CLI commands resolve the upload date and chunk into 90-day windows. Decide per report which is correct (the retention report appears to accept lifetime ranges in one query — verify live).
[...]

PR #140, opened 2026-07-12
feat(#124): implement --output on config + mutation commands; remove auth's dead flag

### LINKED ISSUE #124: --output flag is dead on 6 commands; get-channel-analytics bypasses getOutputFormat

prog893:
Closes #124 (final half — the config-bypass half merged in #133).

## Value proposition

Six commands advertised `--output` and silently ignored it. Now five of them honor it — which turns the mutation commands into scriptable building blocks: `update-video-tags … --yes --output json | jq .newTags` works, and the localization pair returns machine-readable results ready for the #111 batch-localization workflow. The sixth (`auth`) loses the flag: an interactive OAuth flow has no data to format, and advertising a no-op flag is the bug this issue is about.

## Contract

- **`pretty` (default): byte-identical to current behavior.** Previews, colors, confirmation — untouched.
- **Machine formats (`json`/`table`/`text`/`csv`): stdout carries only the result object.** Spinners/success/info already write to stderr; previews become pretty-only; and `confirm()` prompts now go to **stderr** (lib/utils.ts) so a piped stdout stays parseable even when the user forgets `--yes`.
- `config get` keeps its raw-value output for `text`/`pretty` — existing scripts (`config get default.channel`) are unaffected; `json`/`table`/`csv` get `{key, value}` rows.
- Result shapes: update-video → `{videoId, title?, description?, url, dryRun}`; tags → `{videoId, title, previousTags, newTags, addedTags, removedTags, dryRun}`; localization → `{videoId, language, languageName, title, description, action}`.
- `--dry-run` in machine formats emits the proposed change with `dryRun: true` — a preview API for scripts.
- **Breaking (tiny)**: `auth --output x` now errors as an unknown option (it previously did nothing).

## Verification (live on @staqan, zero mutations)

- `config list/get/set --output json` → correct rows; raw `config get` unchanged (`pretty`)
- `update-video --dry-run --output json` and `update-video-tags --dry-run --output json` → clean JSON only on stdout (verified with 2>/dev/null), `addedTags`/`dryRun` correct; pretty dry-run byte-identical
- `auth --output json` → `unknown option`, exit 1
- Localization commands exercised to the API boundary via nonexistent video ID → exit 1, **empty stdout** on failure in machine mode
- No unlisted videos exist on the channel, so the final `videos.update` write wasn't live-fired (public-video writes are off-limits); the post-update emission is the same `formatData` call shape as the verified dry-run path

Docs: `configuration.md` updated (auth flag removed, config semantics noted); the metadata-management docs already listed `--output` for these commands — they were aspirational, now they're true.

## Summary by CodeRabbit

* **New Features**
  * Added configurable machine-readable output formats for configuration, video metadata updates, localization creation/update, and tag updates.
  * Non-`pretty` modes now emit structured, pipeable results (including dry-run details where applicable).
* **Documentation**
  * Updated the configuration command docs to explain `--output` behavior and `get` formatting.
  * Removed the unsupported `--output` flag from authentication command docs.
* **Bug Fixes**
  * Progress indicators and confirmation prompts now write to stderr to keep stdout clean for piped output.

### DIFF
--- a/bin/staqan-yt.ts
+++ b/bin/staqan-yt.ts
@@ -178,11 +178,11 @@ program
     program.help();
   });
 
-// Auth command
+// Auth command. No --output flag: the interactive OAuth flow produces no
+// data to format (the flag was advertised but never read — issue #124).
 program
   .command('auth')
   .description('Authenticate with YouTube API using OAuth 2.0')
-  .option('--output <format>', 'Output format: json, table, text, pretty, csv')
   .option('-v, --verbose', 'Enable verbose output with debug information')
   .action(authCommand);
 
--- a/commands/config.ts
+++ b/commands/config.ts
@@ -1,9 +1,10 @@
 import chalk from 'chalk';
 import * as fs from 'fs/promises';
 import * as path from 'path';
-import { getConfig, getConfigValue, setConfigValue, DEFAULT_LOCK_TIMEOUT_MS } from '../lib/config';
+import { getConfig, getConfigValue, setConfigValue, getOutputFormat, DEFAULT_LOCK_TIMEOUT_MS } from '../lib/config';
 import { success, info, CACHE_DIR } from '../lib/utils';
-import { ConfigKey, CONFIG_KEYS, CONFIG_KEY_HELP } from '../types';
+import { formatData } from '../lib/formatters';
+import { ConfigKey, CONFIG_KEYS, CONFIG_KEY_HELP, OutputFormat } from '../types';
 import { installCompletion, detectShell } from '../lib/completion';
 
 function availableConfigKeysHelp(): string {
@@ -30,6 +31,7 @@ interface ConfigOptions {
   show?: boolean;
   install?: boolean;
   print?: boolean;
+  output?: OutputFormat;
 }
 
 /**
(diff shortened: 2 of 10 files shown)


prog893:
@coderabbitai review

prog893:
Applied 2 of 3 findings in the latest commit:

1. ✅ `update-video-tags` now resolves `--output` before the try block and any API call — verified live: `--output yaml` fails fast with the validation message, exit 1, zero quota spent.
2. ✅ `formatData` got the `never`-based exhaustiveness default.
3. ⏭️ Skipping the shared `printResult(outputFormat, data, prettyRenderFn)` extraction for now: with four call sites whose pretty renderings differ structurally (inline previews vs post-hoc summaries vs spinner-adjacent lines), the helper would need enough parameters to obscure more than it removes. Worth revisiting when #102's structured-returns migration lands — that refactor subsumes this one properly.

prog893:
@coderabbitai review

Issue #105, opened 2026-07-02
lib/language.ts: accept any valid BCP-47 language code (currently hardcoded to en/ja/ru)


prog893:
## Problem

`lib/language.ts` only supports en/ja/ru. Adding a language requires code changes. STAQAN's content is JP-focused but the CLI itself is used in multi-language contexts.

## Current State

Hardcoded language normalization in `lib/language.ts`:
```typescript
// Supports "Japanese", "japanese", "ja", "jpn", "jp" → all normalize to "ja"
```

## Proposed Solution

1. **Move language data to JSON file** at `lib/languages.json` (or similar config location)
2. **Load on startup** with hot-reload support
3. **Allow user override** via `~/.staqan-yt-cli/languages.json` for custom aliases
4. **Use YouTube API** `i18nLanguages.list` endpoint to auto-populate supported languages

## References

- Documented in `staqan/research/staqan-yt-cli-architecture.md` (file 61) as "Potential Issue #5: Language support is hardcoded"

prog893:
**Rescope (2026-07-10 audit reality-check):** the problem is real — only en/ja/ru pass validation, so localizing into, say, Korean or Kazakh is impossible without a code change. But the proposed machinery (JSON config file, hot-reload, per-user alias overrides) is heavy for what this is.

Simpler revised scope (title updated):

1. Accept **any structurally valid BCP-47 code** (`ko`, `kk`, `zh-Hant`, `pt-BR`, …) — YouTube's localization API takes them directly; a simple pattern check catches typos.
2. Keep the existing friendly-name aliases (Japanese → ja etc.) as a static map — extend it opportunistically, no external file.
3. Optionally validate against `i18nLanguages.list` (1 quota unit, cacheable) for a *warning* on codes YouTube doesn't localize — warn, don't block.

Dropped: hot-reload, `~/.staqan-yt-cli/languages.json` overrides.

prog893:
Measured the scope of this against the live channel. It is a hard functional block, not a display limitation, and the channel already has content in languages the localization commands refuse to accept.

## The channel already uses unmapped languages

A sweep of 25 caption tracks across 14 videos returns four languages, one of them outside the mapped set:

| language | tracks | mapped in `lib/language.ts` |
|---|---|---|
| `en` | 12 | yes |
| `ja` | 8 | yes |
| `ru` | 4 | yes |
| `vi` | 1 | **no** |

[...]

PR #148, opened 2026-07-13
feat: add put-caption command (upload caption tracks)

prog893:
## Value proposition

The caption surface was read-only: `list-captions` and `get-caption` exist, but uploading subtitles meant YouTube Studio. `put-caption` completes the workflow — directly relevant to the localization pipeline (upload ja/en/ru subtitle files from the CLI, scriptable via `--output json`). Named `put-caption` per the house `put-*` convention (`put-video-localization`): create-new, API fails if a track with the same language+name exists.

## Command

```bash
staqan-yt put-caption --video-id ID --language ja --file ./subs/ja.srt [--name NAME] [--draft] [--output json]
```

- **`lib/youtube.ts`**: `uploadCaption()` via `captions.insert` with a file stream; requires the `youtube.force-ssl` scope (already granted — same scope `get-caption` uses).
- **Language**: any BCP-47 code passes through (known codes normalized for the display name — the en/ja/ru map limitation is #105's scope, not repeated here).
- **File validation before quota**: existence, regular-file, non-empty — with the supported-formats hint in the error.
- **Output contract per #140**: pretty shows caption ID/track info; machine formats emit `{id, videoId, language, name, isDraft, action}`.
- Registered in bin + help grouping (Engagement) + bash/zsh completions (6 regions, `--file` completes filesystem paths in zsh) + `docs/commands/engagement.md` section.

## Verification

- Help grouping renders; missing/empty file fails fast with exit 1 and **zero API spend**
- Full upload plumbing exercised to the API boundary with a bogus video ID → clean API error, exit 1, no mutation (channel has no unlisted videos and public-video writes are off-limits, so the happy path awaits the next unlisted upload or an explicit go-ahead — the code path difference is only the API's accept vs reject)
- Generated completion scripts contain the new entries and pass `bash -n`/`zsh -n`
- `bun run type-check` ✓ · `bun run lint` ✓ (0 errors, 11 pre-existing warnings) · `bun run build` ✓ · `bun test` 61 pass

Follow-up candidates (not in scope): `update-caption` (captions.update — replace track content by ID) and `delete-caption`. Say the word and I'll file them.

## Summary by CodeRabbit

* **New Features**
  * Added the `put-caption` command for uploading caption or subtitle tracks to videos.
  * Supports language, file, track name, draft uploads, output formatting, and verbose mode.
  * Added human-readable and machine-friendly upload results.
  * Added Bash and Zsh tab-completion support.

* **Documentation**
  * Added command usage, options, examples, and upload behavior details to the engagement documentation.

### DIFF
--- a/bin/staqan-yt.ts
+++ b/bin/staqan-yt.ts
@@ -31,6 +31,7 @@ import listCommentsCommand = require('../commands/list-comments');
 import getChannelCommand = require('../commands/get-channel');
 import listCaptionsCommand = require('../commands/list-captions');
 import getCaptionCommand = require('../commands/get-caption');
+import putCaptionCommand = require('../commands/put-caption');
 import getChannelAnalyticsCommand = require('../commands/get-channel-analytics');
 import getChannelSearchTermsCommand = require('../commands/get-channel-search-terms');
 import listReportTypesCommand = require('../commands/list-report-types');
@@ -449,6 +450,19 @@ program
   .option('--format <format>', 'Caption format: srt, vtt, sbv, scc, ttml, json (default: json)', 'json')
   .option('-v, --verbose', 'Enable verbose output with debug information')
   .action(withHelpWrapper('get-caption', getCaptionCommand));
+
+// Upload a new caption track (PUT - fails if same language+name track exists)
+program
+  .command('put-caption')
+  .description('Upload a new caption track to a video (fails if same language+name track exists)')
+  .requiredOption('--video-id <id>', 'Video ID (must be on your channel)')
+  .requiredOption('--language <lang>', 'Caption language (BCP-47 code, e.g. "ja", "en")')
+  .requiredOption('--file <path>', 'Caption file: srt, vtt, sbv, scc, or ttml')
+  .option('--name <name>', 'Track name shown in the player (default: empty)')
+  .option('--draft', 'Upload as draft (not visible to viewers)')
+  .option('--output <format>', 'Output format: json, table, text, pretty, csv')
+  .option('-v, --verbose', 'Enable verbose output with debug information')
+  .action(withHelpWrapper('put-caption', putCaptionCommand));
 
 // Channel search terms command — top keywords from YouTube Search traffic
 program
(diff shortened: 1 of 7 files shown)


prog893:
@coderabbitai review

prog893:
@coderabbit review

prog893:
I think put may fail to warn about overwrites, but there is no way to replace-overwrite subs then, so maybe --force or --overwrite flag would make sense

prog893:
All three review points addressed in the latest commit:

**@prog893's overwrite gap → `-f, --force`**: when a track with the same language+name exists, `--force` replaces its content via `captions.update` (the only replace mechanism the API offers) instead of failing; without a match it falls through to a normal insert. Result objects carry `replaced: true/false` and `action: "replaced" | "uploaded"`, messaging says *Replaced* vs *Uploaded*, and the pretty output warns when an overwrite happened. Chose `--force` over `--overwrite` for consistency with `fetch-reports --force`. Docs + completions updated.

**CodeRabbit #1 (parseVideoId before spinner)**: hoisted. (Note: `parseVideoId` passes unmatched input through rather than throwing, so the API remains the final validator — the hoist is still the right structure.)

**CodeRabbit #2 (fd leak)**: both the insert and the new update path now retain their read stream and `destroy()` it when the API call throws.

Verified: `--force` exercised to the API boundary (captions.list on a nonexistent video → clean error, exit 1, zero mutation); type-check/lint/build/tests all green. The force-hit replace path shares the same no-unlisted-video caveat as the insert happy path.

prog893:
@coderabbitai review

prog893:
**Post-merge live verification complete** (2026-07-13, on the unlisted test video V_9Yq72iMZM provided by @prog893 — all changes reverted after):

| Path | Result |
|---|---|
| Insert (happy path) | ✅ `ko`/`cli-test` draft track created — `{replaced: false, action: "uploaded"}` |
| Duplicate without `--force` | ✅ API rejects with the same-language+name error, exit 1 (put contract) |
| `--force` replace | ✅ Same caption ID retained, content replaced via `captions.update` — `{replaced: true, action: "replaced"}` |
| Revert | ✅ Test track deleted; video back to its baseline (`ar` auto-track only) |

The full verification matrix from the PR body is now closed — every code path exercised against the live API. (The revert used `captions.delete` directly; a `delete-caption` command remains a follow-up candidate.)

PR #157, opened 2026-07-29
fix(#153): key reporting cache by authenticated channel, validate --channel

### LINKED ISSUE #153: get-report-data: Reporting cache is keyed by user-supplied channel while data comes from the authenticated account
Found by CodeRabbit on #152; pre-existing since the caching pipeline was introduced (declined in-PR as a behavior change out of scope for the #102 extraction).

`fetchReportData` (lib/reports.ts) resolves `--channel` / `default.channel` only to namespace the cache directory. The `jobs.list` / `jobs.reports.list` / download calls always run against the authenticated Reporting API account. Passing a different valid handle therefore writes the authenticated channel's report data under the other channel's cache path — silently poisoning that cache namespace.

Proposed fix: resolve the authenticated channel once (`channels.list({ mine: true })`) and fail loudly when the requested channel doesn't match, or drop the channel parameter entirely and always key by the authenticated channel. Needs a decision on whether any legitimate multi-channel use exists (content-owner delegation is not used; `onBehalfOfContentOwner` is always undefined).

Affects `get-report-data` and `fetch-reports` (same pattern).

prog893:
## Value proposition

The Reporting API has no channel parameter — `jobs.list`, `jobs.reports.list`, `jobs.create`, and `downloadReport` all operate on the authenticated account (`onBehalfOfContentOwner` is always undefined in this CLI). The previous `fetchReportData` / `fetch-reports` flow resolved `--channel` / `default.channel` only to namespace the cache directory, so a wrong handle silently wrote the **authenticated** account's data under the **requested** channel's path. Any `--channel` that didn't match the authed account would poison that cache namespace.

This fix aligns the cache namespace with the actual data source: the authenticated channel. `--channel` / `default.channel` is kept as a validation hook (resolves the handle/ID and fails loudly on mismatch), reserved for future multi-account auth-swap where it will become the credential-selector parameter. Today it surfaces stale `default.channel` config or wrong handles as clear errors instead of silent mislabeling.

## What changed

- **`lib/youtube.ts`** — new `getAuthenticatedChannelId()`: one `channels.list({ part: ['id'], mine: true })` per process, memoized. Single auth context per process → one lookup suffices.
- **`lib/reports.ts:fetchReportData`** — cache namespace = authed channel ID. If `params.channel` is supplied, resolve via `getChannelId` and throw a clear error on mismatch. Dropped the unreachable "No channel specified" error since the authed channel is always available.
- **`commands/fetch-reports.ts`** — same pattern: resolve authed channel for cache namespace, validate `--channel` / `default.channel` against it.
- **`commands/fetch-reports.ts`** — drop now-unused `channelHandle` from `loadCacheIndex(...)` call site (the second `channelHandle?` arg of `loadCacheIndex` is still optional).
- **`docs/commands/reporting-api.md`** — update `--channel` description and the "Required: set default.channel" note for both `get-report-data` and `fetch-reports`. Cache is now keyed by the authenticated channel by construction.

## Why validate instead of ignore

A pure "drop the parameter" approach would let `default.channel` config rot silently until users notice mislabeled data. Validation surfaces the mismatch at the moment of misuse, and the validation hook is exactly what future multi-account auth-swap will need anyway — today's `mine:true` check is tomorrow's "did the swap land on the right channel?" check.

## Cost

- `+1 quota unit` per Reporting command run for `channels.list({ mine: true })`. Memoized per process, so subsequent calls in the same run (e.g. a command that calls `fetchReportData` more than once) cost nothing. Acceptable trade-off for closing a silent data-corruption bug.

## Test plan

- [x] `bun run type-check` — clean
- [x] `bun run lint` — clean (4 pre-existing `any` warnings, unrelated)
- [x] `bun run build` — clean
- [ ] Manual verification of mismatch path: set `default.channel` to a different valid handle and confirm `fetch-reports` / `get-report-data` exit with a clear error rather than silently using the wrong cache path.

Closes #153

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

## Summary by CodeRabbit

- **New Features**
  - Reporting data and caches are now consistently associated with the currently authenticated channel.
  - Added validation to detect when a requested channel does not match the authenticated channel.

- **Bug Fixes**
  - Prevented authenticated data from being stored or retrieved under an incorrect channel.

- **Documentation**
  - Clarified channel option behavior and explained authenticated-channel cache scoping.
  - Documented that channel mismatches may result in an explicit error.

### DIFF
--- a/commands/fetch-reports.ts
+++ b/commands/fetch-reports.ts
@@ -8,7 +8,7 @@ import {
   loadReportMetadata,
   ensureCacheDir,
 } from '../lib/cache';
-import { getChannelId } from '../lib/youtube';
+import { getChannelId, getAuthenticatedChannelId } from '../lib/youtube';
 import { getConfigValue, getLockTimeout } from '../lib/config';
 import { acquireLock, getLockPath } from '../lib/lock';
 import { downloadReport } from '../lib/reports';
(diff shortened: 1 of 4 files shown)


prog893:
@coderabbitai review

prog893:
Addressed all 3 findings from this review (commit d617339):

1. **default.channel fallback restored in fetchReportData** — the data layer now reads `params.channel || await getConfigValue('default.channel')` before validating, so get-report-data catches stale config too (was only caught in fetch-reports).

2. **assertChannelMatchesAuthenticated() helper** — extracted to lib/youtube.ts, called from both lib/reports.ts and commands/fetch-reports.ts. Identical resolution + error message stays in one place.

3. **Promise cache in getAuthenticatedChannelId()** — caches the in-flight Promise (not the resolved value), with .catch() that clears the slot on failure so a later caller can retry. Concurrent callers before first resolve now share one channels.list request.

@coderabbitai resume

prog893:
@coderabbitai resume — all three findings have been addressed in commit d617339 (changes vs prior commit):

1. `params.channel || await getConfigValue('default.channel')` now resolved inside `fetchReportData` (not just `fetch-reports`), so stale `default.channel` is caught by both commands.

2. `assertChannelMatchesAuthenticated(requestedChannel, authenticatedChannelId)` extracted to `lib/youtube.ts`; `lib/reports.ts` and `commands/fetch-reports.ts` both call it.

3. `getAuthenticatedChannelId` now caches the in-flight Promise (with `.catch()` clearing on failure for retry).

The current review re-anchors the same thread IDs to the refactored code — please re-evaluate and flip the verdict.

prog893:
@coderabbitai your three findings from the last review were all implemented in commit d617339:

1. The `default.channel` fallback (`params.channel || await getConfigValue('default.channel')`) is now resolved inside `fetchReportData` in `lib/reports.ts`, so `get-report-data` validates stale config too.
2. The duplicated validation is extracted into `assertChannelMatchesAuthenticated()` in `lib/youtube.ts`, called from both `lib/reports.ts` and `commands/fetch-reports.ts`.
3. `getAuthenticatedChannelId()` now caches the in-flight Promise (clearing it on failure so callers can retry).

Could you re-check these against the current diff and let me know if anything is still outstanding? If they're all resolved, please update the review verdict.

Issue #155, opened 2026-07-17
fetchReportData: availableRange bridges disjoint cache/API coverage; expose gaps in the structured result


prog893:
Found by CodeRabbit on #152; pre-existing behavior moved verbatim from get-report-data (declined in-PR as out of scope for the #102 extraction).

`fetchReportData` (lib/reports.ts) computes `effectiveMinDate`/`effectiveMaxDate` as the min/max over API and cache coverage. Disjoint coverage (e.g. cache holds January, API holds June) is presented as one continuous `availableRange`, and `status: 'ok'` is returned even when interior gaps remain unfilled.

The CLI command mitigates this with its incomplete-data stderr warning (computed from actual row dates), but MCP consumers only see the serialized result and get no gap signal.

Proposed fix: compute uncovered sub-ranges (the data is already in `analyzeCacheCoverage` + fetched windows) and add them to the `ok` result (e.g. `uncoveredRanges: [{startDate, endDate}]`), so both surfaces can present gaps from one source of truth.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #154, opened 2026-07-17
fetchReportData: returns no-reports-yet without checking the local cache


prog893:
Found by CodeRabbit on #152; pre-existing behavior moved verbatim from get-report-data (declined in-PR as a flow restructure out of scope for the #102 extraction).

When `jobs.reports.list` returns zero reports, `fetchReportData` (lib/reports.ts) immediately returns `status: 'no-reports-yet'` ("wait 48h"). If the job is old and all API reports have expired (30–60 days), locally archived data from `fetch-reports` exists but is never consulted — the command claims no data is available while the archive holds it.

Proposed fix: when the API returns zero reports, check `findCachedReports` for the requested range and serve from cache; only return `no-reports-yet` when neither source can cover the request.

Low practical urgency: a live job keeps producing dailies, so the zero-report state realistically occurs only for brand-new jobs (where the message is correct) or long-dormant setups.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #158, opened 2026-08-05
fix(#154,#155): serve reports from cache when API is empty, expose coverage gaps

prog893:
Addresses #154 and #155, the two remaining spin-offs from the #102 extraction. Both live in `fetchReportData`, which is why they ship together. No `Closes:` keyword: GitHub only auto-closes one issue per keyword reliably, so both stay open for manual closing.

## Value proposition

The local archive exists so report data survives YouTube expiring it after 30-60 days. Both bugs undermined that: one made the archive unreachable, the other made incomplete data look complete. Together they meant a user could be told "no data" while holding it, or handed a range with silent holes in it.

## #154: the archive was never consulted before giving up

The zero-reports early return fired the moment the API listed nothing, forty lines above the first `findCachedReports` call:

```ts
if (reports.length === 0) {
  return { status: 'no-reports-yet', ... };   // cache never touched
}
```

For an old job whose API reports have all expired, `fetch-reports` has the data on disk and the command still said "wait 48 hours". The `readyAt` it returned was `jobCreateTime + 48h`, which for such a job is **a timestamp in the past**, so the CLI printed "0 hours remaining" and MCP serialized a stale date.

Now an empty API listing triggers an unbounded archive scan, and `no-reports-yet` is returned only when neither source holds anything. When the API is empty but the archive is not, date bounds are derived from the archive and the run serves entirely from cache.

## #155: disjoint coverage was reported as continuous

`availableRange` is a plain min/max across API and cache. Cache holding January and API holding June produced `{2026-01-01, 2026-06-30}` with `status: 'ok'` and no indication that February through May is a hole.

Added `uncoveredRanges` to the `ok` result, derived from the actual data windows of the reports that actually contributed rows. `availableRange` is now documented as an outer bound, with `uncoveredRanges` as the honest signal.

The CLI's `Incomplete Data` warning now reads that field instead of re-deriving gaps from returned row dates. That old approach had a false-positive: a report can legitimately cover a date and carry no rows because nothing happened that day, and those quiet days were reported as missing data under a message claiming the data "may have expired". The new computation measures what the message actually says, and MCP consumers get the same signal instead of nothing.

## Also fixed: aliasing bug in `mergeDateRanges`

Found while building on it. `[...ranges]` is a shallow copy, so seeding the accumulator with `sorted[0]` and extending `last.end` in place wrote through to the caller's own objects:

```
input before: [{start: 2026-01-01, end: 2026-01-10}, {start: 2026-01-11, end: 2026-01-20}]
input after : [{start: 2026-01-01, end: 2026-01-20}, {start: 2026-01-11, end: 2026-01-20}]
```

`analyzeCacheCoverage` passes the same `{start,end}` objects to `findDateGaps` that it afterwards classifies as fully/partially covered. Merging two adjacent cached reports therefore widened the first report's end date before classification, which pushed a too-wide range into `fullyCovered`, loaded the same cached report twice, and inflated the "Retrieved N cached report(s)" count plus its expiration notices. Row data itself was spared by the existing byte-identical dedup.

## Consolidation

`reports.ts` had reimplemented cache.ts's metadata-first date resolution inline. Both now call the exported `getActualDates`, so the two cannot drift.

## Test plan

- [x] `bun run type-check` clean
- [x] `bun run lint` clean (4 pre-existing `any` warnings, untouched)
- [x] `bun run build` clean
- [x] `bun test` 87 pass / 0 fail (was 68; +19 new in `tests/cache.test.ts`)
- [x] **Regression test verified by reverting the fix**: `mergeDateRanges > does not mutate the caller's range objects` fails without it, passes with it
- [x] New tests cover the #155 disjoint-island case, leading/trailing/single-day gaps, adjacent-is-gapless, and coverage outside the window
- [ ] Live verification against a real expired-report job not performed (requires a dormant job with expired API reports plus a populated archive; the #154 path is exercised by unit-level reasoning and type-checking only)

Refs #154, #155

## Summary by CodeRabbit

- **New Features**
  - Report results now identify explicit date ranges with missing source coverage.
  - Archived local reports can be used when API results are unavailable.
  - Coverage summaries distinguish covered days, gaps, and dates with no activity.
  - Video-specific queries no longer display coverage warnings.

- **Documentation**
  - Updated reporting guidance and tool descriptions to explain coverage boundaries, archived data, and uncovered ranges.

- **Bug Fixes**
  - Improved date-range handling and prevented range merging from altering existing data.

### DIFF
--- a/commands/get-report-data.ts
+++ b/commands/get-report-data.ts
@@ -68,7 +68,7 @@ async function getReportDataCommand(options: ReportDataOptions): Promise<void> {
       return;
     }
 
-    const { rows: filteredData, cachedReports, fetchedReports } = result;
+    const { rows: filteredData, cachedReports, fetchedReports, uncoveredRanges } = result;
     const { startDate: requestedStart, endDate: requestedEnd } = result.requestedRange;
     const { startDate: adjustedStart, endDate: adjustedEnd } = result.adjustedRange;
     const { startDate: effectiveMinDate, endDate: effectiveMaxDate } = result.availableRange;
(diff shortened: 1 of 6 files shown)


prog893:
Both findings were valid and are fixed in 7b7d587.

**1. Archive reports covering a requested subrange (Major)**

Confirmed. `analyzeCacheCoverage` pushes to `fullyCovered` only when `overlap.start === cachedStart && overlap.end === cachedEnd` (lib/cache.ts:569), so "fully covered" actually means "the cached report sits entirely inside the requested range". The reverse containment, one archived report spanning a narrower request, lands in `partiallyCovered` and step 6 never read it. Your Jan 1-31 report / Jan 10-20 query example is exactly right, and the #154 archive-only path made it reachable with no API reports to compensate: empty rows plus a range reported as uncovered.

Step 6 now queries `findCachedReports(channelId, type, adjustedStart, adjustedEnd)` directly and loads every overlap, deduplicated by report ID. Rows are already clamped to the adjusted range downstream, so loading a wider report is safe. `coverage` is still used for the download decision.

**2. Coverage for empty downloaded reports (Major)**

Confirmed. `fetchedWindows` is only appended when `dataMinDate && dataMaxDate` are truthy, so a report covering a period with zero rows contributed nothing and `uncoveredRanges` reported a phantom gap. That is the same false positive this PR set out to remove on the cache side, so it would have been an embarrassing hole.

Added a separate `fetchedCoverage` array, recorded for every successful download: row-date bounds when present, the report's own API window otherwise (matching `getActualDates`' fallback). `uncoveredRanges` now reads it.

I kept it **separate from `fetchedWindows` on purpose** rather than just relaxing the existing push. `fetchedWindows` drives the "fresh downloads supersede cached rows" filter, so letting an empty download claim a window there would discard real cached rows in favour of nothing. Coverage and supersession are genuinely different questions.

Verification: type-check, lint, build clean; 87 tests pass. Note that both fixes are in filesystem-dependent orchestration code, so they are not unit-covered; the new tests in `tests/cache.test.ts` cover the pure date helpers only. Flagging that rather than implying coverage I did not add.

PR #166, opened 2026-08-11
fix(#147): report real caption track state instead of hardcoded false

### LINKED ISSUE #147: list-captions: 'Auto' field reports Yes for manual captions (trackKind: standard, isAutoGenerated: false)
## Bug

`staqan-yt list-captions --video-id <ID>` shows contradictory fields on the same row for manually-uploaded captions:

```
Track Kind: standard
Auto:       Yes
```

For a video where I uploaded captions manually via YouTube Studio (verified byte-for-byte against my source SRT files), every caption row shows `Track Kind: standard` (correct) alongside `Auto: Yes` (incorrect — `isAutoGenerated` should be `false`).

## Reproduction

1. Upload any caption track manually via YouTube Studio (SRT file with hand-crafted timing/transcript).
2. Run `staqan-yt list-captions --video-id <VIDEO_ID>`.
3. Observe `Track Kind: standard` AND `Auto: Yes` on the same row.

## Expected

`Auto: No` for manually-uploaded captions (`isAutoGenerated = false`).

## Likely cause

CLI appears to be reading the wrong field from the YouTube Data API `captions` resource. `isAutoGenerated` is the correct boolean; possibly confusing it with `trackKind === 'ASR'` or some other field.

## Impact

Misleading output caused me (a subagent acting on Tima's behalf) to misdiagnose a video's captions as auto-generated when they were manual, leading to a bad recommendation to 'upload proper captions.' Manual upload had already happened.

## Environment

- staqan-yt-cli v2.0.0 (brew)
- Video tested: `eQICyHf3hsY` (3 caption tracks, all manual)

prog893:
Closes #147.

## Why this matters

`list-captions` was fabricating the part of its answer that decides whether captions actually work. `mapCaptionItem` discarded most of the caption snippet and substituted literals:

| Field | Was | API actually returns |
|---|---|---|
| `isClosedCaptions` | hardcoded `false` | `snippet.isCC` |
| `isLarge` | hardcoded `false` | `snippet.isLarge` |
| `isEasyReader` | hardcoded `false` | `snippet.isEasyReader` |
| `isDraft` | hardcoded `false` | `snippet.isDraft` |
| `languageName` | the language code | resolvable from `language` |
| `status`, `isAutoSynced`, `audioTrackType`, `failureReason`, `name`, `lastUpdated` | dropped | present |

The `isDraft` case is the damaging one. A draft track, a track still `syncing`, and a track whose processing `failed` all appear in the listing but show nothing to viewers. Reporting `isDraft: false` for every track made a track that is not live indistinguishable from one that is: the output does not say "I don't know", it says the wrong thing in the same shape as the right answer.

Every one of these fields arrives in the `part=snippet` response `listCaptions` already requests, so reading them costs no extra quota. The data was fetched and thrown away.

## `serving` does not mean visible

End-to-end verification surfaced a trap. A draft track is reported by the API as `status: "serving"` **and** `isDraft: true` at the same time. Visibility is `status === 'serving' && !isDraft`; a check keyed off `status` alone announces an invisible track as visible. This is now noted at the visibility check, on the `CaptionTrackStatus` type, in the docs, and pinned by a regression test, because it is exactly the kind of thing a later cleanup would collapse into a status-only check.

## Changes

- Populate `isClosedCaptions` (`isCC`), `isLarge`, `isEasyReader`, `isDraft` and `isAutoSynced` from the snippet.
- Add `status` (`serving` / `syncing` / `failed` / `unknown`), `audioTrackType`, `failureReason`, `name` and `lastUpdated` to the caption model. `status` and `audioTrackType` are narrowed against the documented enums rather than cast, and an unrecognized status becomes `unknown` rather than `serving`, since claiming a track is live is the more damaging of the two wrong answers.
- Resolve `languageName` via `getLanguageName`, falling back to the code for languages outside the mapped set (see #105). `en (en)` becomes `English (en)`.
- Surface status and draft state in `table`, `text` and `csv`, and add a `Visibility` line to `pretty` that calls out tracks not currently shown on the video.
- Document the new fields, the visibility rule, and a jq recipe that filters on both fields.

## Verification

**Live sweep**, 25 real caption tracks across 14 videos, all five output formats. Confirms the mapper holds on real data and exercises the `languageName` fallback: tracks in `vi` and `ar` fall back to the bare code (the mapped set is `en`/`ja`/`ru`, see #105), while `en`/`ja`/`ru` resolve to English/Japanese/Russian. 23 of 25 real tracks have an empty `name`, so the `pretty` title falls back correctly.

**Draft round-trip** on an unlisted internal video, since the archive contains no non-serving track and live data alone therefore cannot exercise the path. A draft track was uploaded with `put-caption --draft`, listed, then deleted; the video is back to its original single track.

Same track, on `main`:

```
2. ru (ru)
   Caption ID: AUieDaakrFRpnp_wmmBWrWQlxFWYsDovSeAztPu3SeqU
   Source:     manual (user-uploaded)
```
```json
{ "languageName": "ru", "trackKind": "manual", "isDraft": false }
```

Same track, on this branch:

```
2. Russian (ru)
   Caption ID: AUieDaakrFRpnp_wmmBWrWQlxFWYsDovSeAztPu3SeqU
   Source:     manual (user-uploaded)
   Visibility: draft (not visible to viewers)
   Note:       this track is not currently shown on the video
```
```json
{ "languageName": "Russian", "trackKind": "manual", "isDraft": true, "status": "serving" }
```

`main` reports a genuinely invisible track as an ordinary live one.

`type-check` clean, `lint` clean (4 pre-existing `no-explicit-any` warnings elsewhere), tests 143 -> 153. The `syncing` and `failed` branches stay unit-tested rather than live, since producing them means deliberately corrupting an upload.

## Scope note

The `trackKind` misclassification also named in #147 (`standard` reported as `Auto: Yes`) was already fixed separately in `c11b159`. This PR covers the rest of the resource, which is what kept the issue open.


prog893:
@coderabbitai review

prog893:
@coderabbitai Both findings addressed, one of them partially. Details below.

**`types/index.ts` visibility rule (Minor).** Fixed in `0523ca8`, which landed after the reviewed commit. The `CaptionTrackStatus` doc comment now states that `serving` is necessary but not sufficient, and that visibility is `status === 'serving' && !isDraft`. This came out of end-to-end verification against the live API: uploading a track with `--draft` and listing it back returns `status: "serving"` together with `isDraft: true`, so the two fields really are independent. That is now also pinned by a regression test and documented with a jq recipe filtering on both fields.

**Expanded metadata in every output format (Major).** Taken for `pretty` in `2c0fe8b`, declined for `table` and `text`.

`pretty` was a real gap: closed captions, large text, easy reader, auto-synced timing and non-primary audio tracks were invisible there. It now prints an `Attributes:` line, but only listing the ones actually set, plus `Updated:` when present. All five are false/default on 25 of 25 tracks in a live sweep of this channel, so printing them unconditionally would bury the fields that decide whether the captions work.

`table` and `text` are deliberately curated subsets in this codebase, not complete records. That is the existing convention across commands, not a choice invented here:

| command | `csv` fields | `table` fields | `text` fields |
|---|---|---|---|
| `list-comments` | 12 | 6 | 6 |
| `list-playlists` | 8 | 5 | 5 |
| `list-captions` (this PR) | 16 | 8 | 9 |

`csv` carries the complete record and is where the full field set belongs; it has all 16. `table` is fixed-width terminal output whose ID column alone is 52 characters, so four more columns would push it past any reasonable terminal width and make the status and draft columns, the ones this PR exists to surface, harder to see rather than easier. Expanding `text` and `table` here would make `list-captions` the only command that ignores the convention.

The stated objective is that a track which is not shown to viewers must be distinguishable from one that is. Every format now meets that: `table` and `text` carry `status` and `isDraft`, `pretty` renders the explicit `Visibility` line plus a `Note` when a track is not live, and `json`/`csv` carry everything.

Issue #161, opened 2026-08-09
get-report-data silently truncates piped output at 64KB (--output json/csv)


prog893:
## Summary

`get-report-data --output json` (and `--output csv`) silently truncates at exactly 65536 bytes when stdout is a pipe. Redirecting to a file returns the complete output. Exit code is 0 in both cases, so nothing signals the loss.

This corrupts the machine-readable output path, which is the one scripts consume. `docs/commands/reporting-api.md` documents piping directly:

```bash
staqan-yt get-report-data --type=channel_reach_basic_a1 --output csv | \
  awk -F, 'NR>1 {ctr=$4/$3*100; printf "%s: %.2f%% CTR\n", $2, ctr}'
```

## Reproduction

Any range whose JSON exceeds 64KB. On a channel with a populated archive:

```
$ staqan-yt get-report-data --type=channel_basic_a3 \
    --start-date=2026-04-15 --end-date=2026-04-25 --output json > out.json
$ wc -c < out.json
   83582

$ staqan-yt get-report-data --type=channel_basic_a3 \
    --start-date=2026-04-15 --end-date=2026-04-25 --output json | wc -c
   65536

$ staqan-yt get-report-data ... --output json | python3 -c "import sys,json; json.load(sys.stdin)"
json.decoder.JSONDecodeError: Unterminated string starting at: line 2094 column 5 (char 65522)
```

65536 is the pipe buffer size, so the process is exiting with a partially drained stdout write.

## What it is not

- **Not a runtime flush bug.** A minimal bun script writing 388KB through a pipe completes intact.
- **Not universal across commands.** `list-videos --limit 200 --output json` produces 66136 bytes and survives piping, so it is specific to this path rather than to any output above 64KB.

That narrows it to something `get-report-data` does around the write: it emits stdout from inside the `withSpinner` callback and then performs further stderr writes (expiration notices, incomplete-data warning, row count) before returning.

## Impact

- Silent data loss on the documented scripting path
- Exit code 0, so pipelines cannot detect it
- Affects `json` and `csv` equally, so both machine formats are unsafe above 64KB
- Any aggregation over piped output is wrong, not merely incomplete, because the final record is usually cut mid-token

## Suggested fix

Ensure stdout is fully drained before the process exits. Options, roughly in order of preference:

1. Await the drain explicitly: capture the boolean from `process.stdout.write(...)` and, when false, await the `drain` event before the action resolves.
[...]

PR #80, opened 2026-06-19
fix: validate --format in get-caption (#69)

prog893:
## Summary
- Adds a whitelist check for `--format` in `get-caption` before any API call
- Invalid values now exit immediately with a clear error message instead of producing a misleading `caption-not-found` API error
- Corrects the valid format list: replaces `srv2` (not in YouTube API docs) with `scc`; keeps `json` as the CLI-level "return raw" default

## Valid formats (per [YouTube API docs](https://developers.google.com/youtube/v3/docs/captions/download))
`srt`, `vtt`, `sbv`, `scc`, `ttml`, `json`

## Test plan
- [ ] `staqan-yt get-caption --caption-id <id> --format invalid` → exits with `Invalid format 'invalid'. Valid: srt, vtt, sbv, scc, ttml, json`
- [ ] `staqan-yt get-caption --caption-id <id> --format srt` → proceeds normally
- [ ] `staqan-yt get-caption --caption-id <id>` (no format) → defaults to `json`, proceeds normally

Closes #69

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added `get-caption --format` validation against the supported caption formats (srt, vtt, sbv, scc, ttml, json), returning an error for unsupported values.
* **Bug Fixes**
  * Improved caption download error messaging for restricted caption permissions, clarifying the limitation when captions aren’t available from the authenticated channel.
* **Documentation**
  * Updated `get-caption` docs, examples, and shell completion suggestions to replace `srv2` with `scc`.
* **Chores**
  * Kept caption format metadata and completion lists consistent across the CLI.

PR #168, opened 2026-08-11
fix(#167): send tfmt so get-caption --format actually converts

### LINKED ISSUE #167: get-caption: --format is validated then ignored; every format returns identical bytes
## Bug

`get-caption --format` is validated client-side and then discarded. `lib/youtube.ts` `downloadCaption()` never passes `tfmt` to `captions.download`, so every value of `--format` returns byte-identical content: whatever format the track was originally uploaded in.

```ts
const response = await youtube.captions.download({
  id: captionId,               // no tfmt
}, { responseType: 'text' });

if (format === 'json') return content;
// For other formats, YouTube returns the format that was uploaded
return content;                // format never used
```

The `format` parameter is accepted, checked against `CAPTION_FORMATS`, threaded through the call, and then never read.

## Reproduction

Four downloads of the same track, differing only in `--format`:

```
json    8698 bytes  md5=7f488731024b02366583bbf7d6736bb9
srt     8698 bytes  md5=7f488731024b02366583bbf7d6736bb9
vtt     8698 bytes  md5=7f488731024b02366583bbf7d6736bb9
sbv     8698 bytes  md5=7f488731024b02366583bbf7d6736bb9
```

`--format vtt` returns SRT: comma decimal separators, numeric cue indices, and no `WEBVTT` header.

```
1
00:00:13,847 --> 00:00:15,147
So we're gonna start doing this.
```

Exit status is 0 throughout, so nothing signals that the requested conversion did not happen.

## Impact

The documented workflows do not work.

`docs/commands/engagement.md` recommends `--format vtt > captions.vtt`. That writes SRT bytes into a `.vtt` file, which web players reject. The failure surfaces later, in another tool, with no indication of where the wrong content came from.

The same doc describes `json` as "Structured JSON with timing (default)" and builds three recipes on it (extract plain text, word cloud, find mentions). All three fail, because the content is not JSON:

```
$ staqan-yt get-caption --caption-id <id> --format json | jq -r '.[].text'
jq: error (at <stdin>:1): Cannot iterate over number (1)
jq: parse error: Expected string key before ':' at line 2, column 3
```

## The API supports this

`captions.download` takes `tfmt`. Passing it produces genuinely different output, verified live against the same track:

| tfmt | result |
|---|---|
| `srt` | 8236 bytes, SRT |
| `vtt` | 7988 bytes, starts with `WEBVTT` |
| `sbv` | 7334 bytes, `0:00:13.847,0:00:15.147` |
| `ttml` | 11685 bytes, `<?xml version="1.0" encoding="utf-8" ?>` |
| `scc` | rejected, HTTP 404 `The caption track could not be found` |
| `json` | rejected, HTTP 400 `The caption track data could not be converted` |

## Fix

Pass `tfmt` for the four formats the API actually converts to, and reconcile `CAPTION_FORMATS` with that list. Two declared formats are not real `tfmt` values:

- `scc` is rejected outright and should be dropped or documented as unsupported.
- `json`, the current default, is not a `tfmt` value either. Today it "works" only as a side effect of `tfmt` never being sent, which returns the track's original format. The honest spelling of that behavior is `raw` or `original`, keeping `json` as an alias so existing scripts keep running, since the default cannot simply change meaning.

A naive fix that forwards `tfmt: format` unconditionally will break the default, because `tfmt=json` is a 400.

Docs need updating in the same change: the format table, the three jq recipes, and the "Caption ID Format" section, which claims IDs look like `<language_code>.<video_id>` when real ones are opaque (`AUieDaZbd2DZ...`).

## Related

Same class as #147: a value is fetched or accepted, quietly discarded, and replaced with something that has the shape of a correct answer. #80 added client-side validation for `--format` without making the flag do anything, which is why this reads as supported.

## Environment

Reproduced against the live API on the authenticated channel at `9d38272` (main).

prog893:
Closes #167.

## The bug

`downloadCaption()` never passed `tfmt` to `captions.download`, so every value of `--format` returned byte-identical content: whatever format the track was uploaded in. `--format vtt` returned SRT. The flag was validated, threaded through the call, and then never read.

Exit status was 0 throughout. The documented `--format vtt > captions.vtt` recipe wrote SRT bytes into a `.vtt` file, and the failure surfaced later in a different tool with nothing pointing back at this command.

## Verification

One download per format, same track, live API:

| `--format` | before | after | first line |
|---|---|---|---|
| `raw` | 8698, md5 `7f488731` | 8697, md5 `4e322db6` | `1` |
| `srt` | 8698, md5 `7f488731` | 8238, md5 `8fa29c79` | `1` |
| `vtt` | 8698, md5 `7f488731` | 7990, md5 `81ed25e4` | `WEBVTT` |
| `sbv` | 8698, md5 `7f488731` | 7336, md5 `1293b6c2` | `0:00:13.847,0:00:15.147` |
| `ttml` | n/a | 11687, md5 `0b156114` | `<?xml version="1.0" encoding="utf-8" ?>` |
| `json` | 8698, md5 `7f488731` | 8697, md5 `4e322db6` | alias, byte-identical to `raw` |

Before, one md5 for everything. After, each format is distinct and correct. The no-flag default is byte-identical to `raw`, so existing behavior is unchanged.

## Format list reconciled with the API

Two advertised values were never real `tfmt` values:

- **`scc`** is rejected with HTTP 404 on every track tried (manual, ASR, two languages). Now declined client-side naming that reason, rather than listed as merely invalid.
- **`json`** is an HTTP 400 as a `tfmt` value. It was the default and only "worked" because `tfmt` was never sent, which returns the original format. That behavior is now spelled `raw`, with `json` kept as an alias so existing scripts keep running.

A fix that forwards `tfmt: format` unconditionally would break the default, which is why the mapping is explicit.

## Also in this PR

**`console.log` to `writeStdout` for the content write.** Redirecting to a file is this command's primary use, and `console.log` truncates piped output at 65536 bytes with exit 0 (#161). This is **latent, not active**: the largest caption on the channel is 11.7KB, so no current track crosses the threshold, but ttml on a long video would. The mechanism is already covered by the #161 regression guard, so no duplicate test is added.

**`GetCaptionOptions.format` is now `string`, not `CaptionFormat`.** Commander does not validate it, and typing it as the narrow union hid retired values like `scc` from the client-side check entirely.

## Docs

Every `get-caption` recipe in `docs/commands/engagement.md` was broken, independently of the `tfmt` bug. All are rewritten and run against the live API:

- Three `jq` recipes parsed output that was never JSON.
- One `xargs` recipe appended the ID *after* `--format json`, so `--caption-id` consumed the `--format` flag as its value.
- The multi-track downloader derived the language with `cut -d. -f1` from an ID format that does not exist. Real IDs are opaque, so every track of a video collided into one filename. It now takes the language from the listing, and includes `trackKind`, since a video can carry a manual and an auto-generated track in the same language.
- The "Caption ID Format" section documented that same fictional `<language_code>.<video_id>` shape.

Recipes verified end to end on a video with three tracks:

```
en-automatic.srt  10116 bytes
en-manual.srt      8238 bytes
ja-manual.srt      8892 bytes
```

Three distinct files, which is exactly what the old `cut -d. -f1` version could not produce.

`type-check` clean, `lint` clean (4 pre-existing `no-explicit-any` warnings elsewhere), tests 153 -> 159.

## Related

Same class as #147: a value is fetched or accepted, quietly discarded, and replaced with something shaped like a correct answer. #80 added the client-side `--format` validation without making the flag do anything, which is why it read as supported.

## Summary by CodeRabbit

- **New Features**
  - Caption downloads now default to the original (`raw`) format.
  - Added support for SRT, VTT, SBV, and TTML conversions.
  - `json` remains supported as an alias for `raw`.
  - Improved format validation, caption output preservation, and handling of unsupported SCC downloads.

- **Documentation**
  - Updated caption download guidance, examples, filtering workflows, and multi-caption instructions.
  - Refreshed shell completion options to match supported formats.

### DIFF
--- a/bin/staqan-yt.ts
+++ b/bin/staqan-yt.ts
@@ -447,7 +447,7 @@ program
   .command('get-caption')
   .description('Download caption content to stdout (get caption ID from list-captions)')
   .requiredOption('--caption-id <id>', 'Caption ID')
-  .option('--format <format>', 'Caption format: srt, vtt, sbv, scc, ttml, json (default: json)', 'json')
+  .option('--format <format>', 'Caption format: raw, srt, vtt, sbv, ttml (default: raw)', 'raw')
   .option('-v, --verbose', 'Enable verbose output with debug information')
   .action(withHelpWrapper('get-caption', getCaptionCommand));
 
--- a/commands/get-caption.ts
+++ b/commands/get-caption.ts
@@ -1,6 +1,6 @@
 import { downloadCaption } from '../lib/youtube';
-import { debug, initCommand, createSpinner } from '../lib/utils';
-import { GetCaptionOptions, CAPTION_FORMATS } from '../types';
+import { debug, initCommand, createSpinner, writeStdout } from '../lib/utils';
+import { GetCaptionOptions, CAPTION_FORMATS, CAPTION_FORMAT_ALIASES, CaptionFormat } from '../types';
 
 async function getCaptionCommand(options: GetCaptionOptions): Promise<void> {
   initCommand(options);
@@ -11,13 +11,25 @@ async function getCaptionCommand(options: GetCaptionOptions): Promise<void> {
     throw new Error('Required: --caption-id');
   }
 
-  if (options.format && !(CAPTION_FORMATS as readonly string[]).includes(options.format)) {
+  // `scc` used to be advertised but the API rejects it with HTTP 404 on every
+  // track tried (manual, ASR, several languages), so name the reason instead
+  // of listing it as merely invalid.
+  if (options.format === 'scc') {
+    throw new Error(
+      "Format 'scc' is not supported: the YouTube API rejects scc conversion " +
+      'for caption tracks (HTTP 404).\n' +
+      `Valid formats: ${CAPTION_FORMATS.join(', ')}`
+    );
+  }
+
+  const requested = options.format || 'raw';
+  const format: CaptionFormat = CAPTION_FORMAT_ALIASES[requested] ?? (requested as CaptionFormat);
+
+  if (!(CAPTION_FORMATS as readonly string[]).includes(format)) {
     throw new Error(`Invalid format '${options.format}'. Valid: ${CAPTION_FORMATS.join(', ')}`);
   }
 
   // Note: For caption metadata, use list-captions --video-id <videoId>
-  // This command focuses on downloading caption content
-  const format = options.format || 'json';
   const spinner = createSpinner(`Downloading caption (${format})...`).start();
 
   try {
@@ -26,8 +38,11 @@ async function getCaptionCommand(options: GetCaptionOptions): Promise<void> {
 
     spinner.succeed('Caption downloaded');
 
-    // Output caption content to stdout (allows redirection)
-    console.log(content);
+    // Must be writeStdout, not console.log: redirecting to a file is this
+    // command's primary use, and console.log silently truncates piped output
+    // at 65536 bytes with exit 0 (issue #161). Caption files cross that
+    // easily, ttml most of all.
+    await writeStdout(content.endsWith('\n') ? content : content + '\n');
   } catch (err) {
     spinner.fail('Failed to download caption');
     const errMessage = (err as Error).message;
--- a/commands/get-caption.ts
+++ b/commands/get-caption.ts
@@ -38,10 +38,16 @@ async function getCaptionCommand(options: GetCaptionOptions): Promise<void> {
 
     spinner.succeed('Caption downloaded');
 
+    // The contract here is byte fidelity, not newline normalization: this
+    // output is a caption file. srt, vtt and sbv end with two newlines, and
+    // that trailing blank line terminates the final cue, so collapsing it to
+    // one would corrupt the format. The conditional only guards against
+    // leaving a shell prompt mid-line if a format ever returns no trailing
+    // newline; every format observed live returns at least one.
+    //
     // Must be writeStdout, not console.log: redirecting to a file is this
     // command's primary use, and console.log silently truncates piped output
-    // at 65536 bytes with exit 0 (issue #161). Caption files cross that
-    // easily, ttml most of all.
+    // at 65536 bytes with exit 0 (issue #161). ttml on a long video crosses it.
     await writeStdout(content.endsWith('\n') ? content : content + '\n');
   } catch (err) {
     spinner.fail('Failed to download caption');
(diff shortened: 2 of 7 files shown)


prog893:
@coderabbitai All three addressed in `6aaed0d`, one of them by documenting the intent rather than changing the behavior.

**MD028 and MD040.** Both fixed. The two format blockquotes are now one, and the caption ID fence is tagged `text`.

**Trailing line endings.** Declining the normalization, because the premise is inverted: there is no one-trailing-newline contract here, and introducing one would corrupt the output. This command writes a caption file, so the contract is byte fidelity with the API response.

Measured live on a single track, counting trailing newlines on the actual bytes:

| `--format` | trailing newlines | final bytes |
|---|---|---|
| `raw` | 1 | `0d 0a` (CRLF) |
| `srt` | 2 | `0a 0a` |
| `vtt` | 2 | `0a 0a` |
| `sbv` | 2 | `0a 0a` |
| `ttml` | 1 | `3e 0a` |

The second newline on `srt`, `vtt` and `sbv` is not stray whitespace. It is the blank line that terminates the final cue, and it is part of those formats. Collapsing it would produce files whose last cue is unterminated. `raw` additionally ends `0d 0a`, because the track was uploaded with CRLF line endings; a normalization pass would rewrite those too, silently changing a file the API returned verbatim.

The conditional therefore is not a partial normalization. It appends a newline only when the response has none at all, purely so a terminal prompt is not left mid-line, and no format observed live ever hits that branch. The call site now says exactly this, so it does not read as an oversight.

Worth noting the byte counts in the PR body are byte counts, not string lengths. An earlier probe of mine reported `srt` as 8236 against the CLI's 8238; the difference was `Ø` and `ñ` encoding to two UTF-8 bytes each, not a discrepancy in what the command writes. The CLI output is byte-identical to the API response.

prog893:
@coderabbitai All three findings from the first review are now resolved: MD028 and MD040 fixed in 6aaed0d, and the trailing-line-ending finding withdrawn above. The PR-level verdict is still CHANGES_REQUESTED from the first pass, which no longer reflects any open finding. Please re-evaluate and post an updated review.

prog893:
@coderabbitai Requesting an explicit sign-off review on this PR.

Status of your three findings from the first pass:

1. **MD028** (blank line between blockquotes, `docs/commands/engagement.md`) - fixed in `6aaed0d`.
2. **MD040** (untagged code fence for the caption ID) - fixed in `6aaed0d`, now ```text.
3. **Trailing line endings** (`commands/get-caption.ts`) - you withdrew this one after the measurements, confirming byte fidelity is the correct contract and that normalizing would corrupt final-cue termination in srt/vtt/sbv and rewrite CRLF in raw.

So nothing is outstanding, but the PR-level verdict is still CHANGES_REQUESTED from that first pass, which no longer matches any open finding. Your later response came through as a COMMENTED review, which does not supersede it.

Please post an updated review so the verdict reflects the current state. If anything above is still open from your side, say which and I will fix it rather than merge past it.

For completeness, CI on the head commit: `checks` SUCCESS, `claude-review` SUCCESS, mergeable state CLEAN.

PR #169, opened 2026-08-14
fix(#105): accept any BCP-47 language tag

### LINKED ISSUE #105: lib/language.ts: accept any valid BCP-47 language code (currently hardcoded to en/ja/ru)

prog893:
Closes #105.

## Not cosmetic

`normalizeLanguage` matched a hardcoded alias table for `en`/`ja`/`ru` and returned null for everything else, and null is a **throw** at three call sites:

```
$ staqan-yt put-video-localization --video-id <id> --language ko --title x --description y
✗ Invalid language: ko                                    (exit 1)

$ staqan-yt get-video-localizations --video-ids <id> --languages vi
✗ No valid languages provided in filter                   (exit 1)
```

`update-video-localization` took the same path. The channel already publishes caption tracks in `vi` and `ar`, so it could not localize into languages it was already serving. The docs meanwhile advertised `es`, `de` and `fr` as "common language codes", none of which the code accepted.

## Validation via Intl

- `Intl.getCanonicalLocales` rejects malformed input and canonicalizes the rest: `EN` to `en`, `zh-hans` to `zh-Hans`.
- `Intl.DisplayNames` with `fallback: 'none'` returns undefined for a well-formed tag naming no language. That is the part that matters: BCP-47 syntax alone accepts any 2-8 letter subtag, so canonicalization on its own would happily pass `bogus` and `xx`.

| input | before | after |
|---|---|---|
| `ko`, `vi`, `ar`, `es`, `de`, `fr` | rejected | accepted, named |
| `zh-hans` | rejected | `zh-Hans`, named as Chinese |
| `pt-br` | rejected | `pt-BR`, named as Portuguese |
| `EN` | rejected | `en` |
| `tlh` | rejected | `tlh`, "Klingon" |
| `bogus`, `xx`, `klingon` | rejected | still rejected |
| `123`, `x`, `''` | rejected | still rejected |

The alias table is kept and checked **first**, for correctness as much as convenience: `jp` is a well-formed tag and a region subtag, but not a language code. Canonicalization alone would accept it and then fail at the API.

`getLanguageName` now names any tag, with the alias table winning for `en`/`ja`/`ru` so those stay stable across ICU versions.

`getSupportedLanguages` became misleading once any tag is accepted, so it is renamed `getAliasedLanguages` with the old name kept as a deprecated alias.

## Knock-on fix to list-captions

`list-captions` resolves `languageName` through the same helper, so the channel's real `vi` and `ar` tracks displayed as `vi (vi)` and `ar (ar)`. Now:

```
1. Vietnamese (vi)
   Caption ID: AUieDaaC9YNO10zMR56eDUT9moi4DvaDwpXzhsrp2K6PUADIuTY
   Source:     automatic (YouTube-generated)
   Visibility: serving (visible to viewers)
```

The #147 test that pinned the old fallback is updated, which is the intended direction: it carried a note saying this expectation must flip when #105 lands.

## Verification

Read paths, live: `--languages vi` and `--languages es,de,fr` now exit 0 where they exited 1; `bogus` is still rejected, with a message that names the expected format.

Write paths, live: a `ko` localization was created, read back, and updated through `put-video-localization` / `get-video-localizations` / `update-video-localization`, then the video was restored to its **exact** original state (`diff` clean against a full pre-capture of `snippet` + `localizations`). This was done on an unlisted internal-use video, never a published one.

One note from that round-trip, since it is a real trap: an incomplete `snippet` on `videos.update` causes YouTube to re-derive omitted fields. A first restore attempt that sent only title/description/categoryId/tags/defaultLanguage silently flipped `defaultAudioLanguage` from `en` to `en-US`. The CLI does not have this bug: `putVideoLocalization` spreads the full `...video.snippet`. It was my restore script, and the field was set back.

## Portability note, caught by CI

The first push passed locally and failed CI. CLDR display strings differ by ICU version and platform: macOS names `zh-Hans` "Chinese, Simplified", Linux names it "Simplified Chinese". The test asserted one platform's phrasing.

Fixed in `155e6ac` by asserting what the code guarantees for script and region variants (a resolved name that is not the raw tag, naming the right language) rather than an exact string. Single-word names like Korean and Vietnamese pass identically on both platforms and are unchanged.

This is the same reasoning that already justified the alias table pinning `en`/`ja`/`ru`; it just had not been carried into the test. Display names are for humans to read, never for callers to compare against, and that is now stated at `getLanguageName`.

`type-check` clean, `lint` clean (4 pre-existing `no-explicit-any` warnings elsewhere), tests 159 -> 163.

## Summary by CodeRabbit

* **New Features**
  * Localization commands now support recognized BCP-47 language tags, including regional and extended variants.
  * Language tags are automatically canonicalized and displayed with human-readable names.
  * Existing channel-language aliases remain supported for compatibility.
* **Bug Fixes**
  * Invalid or unrecognized language inputs are rejected earlier with clearer guidance and examples.
  * Caption language names now resolve correctly for additional languages, including Korean and Vietnamese.
* **Documentation**
  * Updated localization command documentation to describe supported language tags, aliases, and validation behavior.

### DIFF
--- a/docs/commands/metadata-management.md
+++ b/docs/commands/metadata-management.md
@@ -104,10 +104,25 @@ staqan-yt get-video-localization --video-id dQw4w9WgXcQ --language ja --output j
 
 ### Language Specification
 
-Accepts any of these formats:
-- Language code: `ja`, `en`, `ru`
-- Language name: `Japanese`, `English`, `Russian`
-- Full locale: `ja-JP`, `en-US`
+Any BCP-47 language tag is accepted:
+
+- Language code: `ja`, `ko`, `vi`, `ar`, `es`
+- Script or region subtags: `zh-Hans`, `pt-BR`, `en-US`
+- Language name, for the channel's three main languages only: `Japanese`,
+  `English`, `Russian` (plus short forms like `jp`, `eng`, `rus`)
+
+Tags are canonicalized, so `EN` becomes `en` and `zh-hans` becomes `zh-Hans`.
+
+A tag that is well-formed but names no actual language is rejected up front:
+
+```bash
+$ staqan-yt put-video-localization --video-id VIDEO_ID --language bogus ...
+✗ Invalid language: bogus. Pass a BCP-47 tag (ja, ko, zh-Hans, pt-BR) or one of the aliases: en, ja, ru.
+```
+
+> Before v2.1.x these commands accepted only `en`, `ja` and `ru` and rejected
+> everything else with `Invalid language`, which blocked localizing into
+> languages a channel might already publish captions in.
 
 ### What is a Localization?
 
--- a/lib/language.ts
+++ b/lib/language.ts
@@ -32,6 +32,33 @@ function canonicalizeTag(tag: string): string | null {
   }
 }
 
+/**
+ * Display name for an already-canonical tag, or null when it names no
+ * language. Never throws.
+ *
+ * `of()` is stricter than `getCanonicalLocales`: it raises RangeError
+ * ("argument is not a language id") for extension and private-use subtags that
+ * canonicalize perfectly well, including legitimate tags like
+ * `de-DE-u-co-phonebk` and `en-x-private`. Those still name a real language, so
+ * the extensions are dropped and the base subtag is named instead, rather than
+ * leaking an ICU error message for valid input.
+ */
+function resolveDisplayName(canonical: string): string | null {
+  try {
+    return displayNames.of(canonical) ?? null;
+  } catch {
+    const base = canonical.split('-')[0];
+    if (!base || base === canonical) {
+      return null;
+    }
+    try {
+      return displayNames.of(base) ?? null;
+    } catch {
+      return null;
+    }
+  }
+}
+
 const LANGUAGE_MAP: LanguageMap = {
   'en': {
     code: 'en',
@@ -87,7 +114,7 @@ function normalizeLanguage(input: string | undefined | null): string | null {
 
   // Well-formed but not a language the platform recognizes. Passing it through
   // would trade a clear client-side error for an opaque one from the API.
-  if (!displayNames.of(canonical)) {
+  if (!resolveDisplayName(canonical)) {
     return null;
   }
 
@@ -139,7 +166,7 @@ function getLanguageName(code: string): string | null {
     return null;
   }
 
-  return displayNames.of(canonical) ?? null;
+  return resolveDisplayName(canonical);
 }
 
 /**
--- a/lib/language.ts
+++ b/lib/language.ts
@@ -1,9 +1,36 @@
 /**
- * Language mapping and normalization utility for YouTube video localizations
- * Supports English, Japanese, and Russian with case-insensitive matching
+ * Language normalization for YouTube localizations and captions.
+ *
+ * Accepts any BCP-47 language tag (issue #105). The alias table below is a
+ * convenience layer for the channel's three main languages, so `japanese`,
+ * `jpn` and `jp` all resolve to `ja`; everything else is validated and named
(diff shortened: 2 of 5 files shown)


prog893:
@coderabbitai review

prog893:
@coderabbitai review

prog893:
@coderabbitai review

prog893:
@coderabbitai Valid finding, fixed in `b7afb8e`, and it reaches further than the reported case.

`Intl.DisplayNames.of()` raises RangeError for **any** extension or private-use subtag that survives `Intl.getCanonicalLocales`, not just private-use:

| tag | canonicalizes | `of()` |
|---|---|---|
| `en-x-private` | `en-x-private` | RangeError |
| `en-US-x-foo` | `en-US-x-foo` | RangeError |
| `de-DE-u-co-phonebk` | `de-DE-u-co-phonebk` | RangeError |
| `en-a-bbb` | `en-a-bbb` | RangeError |
| `zh-Hans-CN` | `zh-Hans-CN` | resolves |
| `x-private` | rejected earlier | not reached |

`de-DE-u-co-phonebk` is the one that matters: a Unicode collation extension is ordinary valid BCP-47, not an edge case someone has to go looking for.

One correction to the severity. This was not an uncaught crash. The RangeError was caught by the command's error handling, so it surfaced as a leaked ICU string with a correct exit code:

```
$ staqan-yt get-video-localizations --video-ids ID --languages de-DE-u-co-phonebk
✗ argument is not a language id                        (exit 1)
```

Bad output for valid input, rather than an availability problem. Still worth fixing, which is why it is fixed.

Both call sites now route through a single `resolveDisplayName` helper that never throws. These tags do name a real language, so on RangeError the extensions are dropped and the base subtag is named: `de-DE-u-co-phonebk` is German, `en-x-private` is English. The tag itself is preserved canonicalized; only the display lookup falls back.

The validity check does not weaken: `bogus-x-private` still resolves to nothing and is still rejected, now with the intended message rather than the ICU one.

```
$ staqan-yt get-video-localizations --video-ids ID --languages de-DE-u-co-phonebk
Localizations for video: ...                           (exit 0)

$ staqan-yt get-video-localizations --video-ids ID --languages bogus-x-private
✗ No valid languages in filter: bogus-x-private. Pass BCP-47 tags (ja, ko, zh-Hans, pt-BR).
```

Regression tests added for `normalizeLanguage` and `getLanguageName` on `en-x-private`, `de-DE-u-co-phonebk` and `en-US-x-foo`, plus the `bogus-x-private` rejection. Those name assertions match on `/english/i` and `/german/i` rather than exact strings, since CLDR wording varies by platform (that is what broke CI on the previous push). Tests 163 to 166.

prog893:
@coderabbitai review

Repository access should be working again now. Commit `b7afb8e` is on the branch with the `resolveDisplayName` helper in `lib/language.ts` and the extension-subtag regression tests in `tests/language.test.ts`. CI `checks` is green on it.

PR #172, opened 2026-08-19
docs: record the merge gate and the E2E proof requirement

prog893:
Two working rules were being followed without being written down, so they got rediscovered by trial and error on every PR. This writes them down.

## Merge gate

CodeRabbit approval plus green CI authorizes a merge; no separate human sign-off. The parts worth documenting are the ones that actually cost time:

- **Green checks are not approval.** A PR can have every check passing and still carry a blocking `CHANGES_REQUESTED`.
- **A `COMMENTED` follow-up does not supersede a `CHANGES_REQUESTED`**, and bare `@coderabbitai review` kicks usually no-op. Asking for an explicit verdict, listing each finding and its status, does work. This was learned twice, on #168 and #169.
- **Total silence with no acknowledgement means quota**, not a stuck review.
- **Findings can be wrong.** Disagreeing with measurements or with the codebase convention gets them withdrawn, and beats complying with a change that makes the code worse. Both happened on #168: a byte-normalization request withdrawn once the real API bytes were shown, and an output-format request withdrawn against a table of what comparable commands actually do.

## E2E proof

Unit tests encode what the author believed while writing the fix, so a fix built on a wrong belief ships with tests that agree with it. The requirement is three-part: capture the broken behavior on `main` **before** merging, re-run the **identical** commands after, and prove no regressions with both positive controls (things that worked still work) and negative controls (invalid input is still rejected).

Evidence should be exit codes, byte counts and checksums. "`--format` works now" is weak; `distinct md5 across 5 formats: before 1, after 5` is not.

Also covered:

- **Trial-merge locally when two PRs touch the same files.** Each PR's CI tested it against the `main` it branched from, never against the other. #168 and #169 overlapped in `lib/youtube.ts` and `tests/captions.test.ts`.
- **Never test mutating commands against published videos.** Unlisted only, capture the full prior state, restore, and `diff` to prove the restore was exact.
- **The `videos.update` trap**: fields omitted from the payload get re-derived by YouTube rather than left alone. A partial restore silently flipped `defaultAudioLanguage` from `en` to `en-US`.

## Structure

Detail goes in `docs/development/e2e-and-merge-gate.md` with a summary in CLAUDE.md, following that file's own instruction to keep tutorials out of the manifesto and link to guides instead. CLAUDE.md grows by 36 lines rather than 66.

Note: CLAUDE.md is 492 lines against its own "keep under 400" rule. It was already 456 before this change, so the overrun predates it, but it is worth a trim at some point.

Also corrects the development guide count in CLAUDE.md, which read 11 while there were already 12.

Docs only. No code, no behavior change.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for the CodeRabbit merge-gate workflow, including review verdicts, CI requirements, and rate-limit handling.
  * Documented end-to-end proof requirements using exit codes, byte counts, checksums, and before/after comparisons.
  * Added recommendations for trial merges and safely testing commands that modify published content.
  * Added an objective example for validating caption-format output.
  * Updated the development documentation index and guide structure.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -371,6 +371,10 @@ staqan-yt-cli/
 - **A `COMMENTED` follow-up does not clear a `CHANGES_REQUESTED`.** Bare
   re-kicks usually no-op; ask for an explicit verdict listing each finding.
 - **Total silence means quota**, not a stuck review.
+- **Never kick a rate-limited PR on a guess.** Compute
+  `available_at = <rate-limit comment>.updated_at + N minutes` (the value is
+  relative to `updated_at`, not `created_at`, and does not tick down). Wait,
+  then kick **once**. Do not poll a PR that is rate limited.
 
 **Every fix needs an E2E proof, not just unit tests:**
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -362,14 +362,19 @@ staqan-yt-cli/
 `APPROVED` review and CI is green. No separate human sign-off is required.
 
 - **Never merge on `CHANGES_REQUESTED`.** Green checks are not approval:
+
   ```bash
   gh pr view <N> --repo prog893/staqan-yt-cli --json reviewDecision,reviews \
     --jq '{decision:.reviewDecision, reviews:[.reviews[]|.state]}'
   ```
+
 - **Argue with it when it is wrong**, using measurements or the codebase
   convention. It verifies and withdraws findings that do not hold up.
 - **A `COMMENTED` follow-up does not clear a `CHANGES_REQUESTED`.** Bare
   re-kicks usually no-op; ask for an explicit verdict listing each finding.
+- **Pick the right command**: `review` is incremental and skips seen commits,
+  `resume` only un-pauses, `full review` forces a re-examination. A rate-limited
+  first attempt marks commits as seen, so only `full review` recovers it.
 - **Total silence means quota**, not a stuck review.
 - **Never kick a rate-limited PR on a guess.** Compute
   `available_at = <rate-limit comment>.updated_at + N minutes` (the value is
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -156,7 +156,7 @@ staqan-yt-cli/
 ├── CONTRIBUTING.md           # Contributor guide
 ├── CLAUDE.md                 # This file - AI development instructions
 └── docs/
-    ├── development/          # Development guides (11 guides)
+    ├── development/          # Development guides (13 guides)
     ├── commands/             # Command reference
     └── *.md                  # Setup, troubleshooting, etc.
 ```
@@ -309,6 +309,7 @@ staqan-yt-cli/
 
 ### Maintenance & Operations
 - **[Git Workflow Guide](docs/development/git-workflow.md)** - Branch strategy, commits, releases, and protection
+- **[E2E Proof and the Merge Gate](docs/development/e2e-and-merge-gate.md)** - Merge gate, E2E proof, and safe live testing
 - **[Maintenance Guide](docs/development/maintenance-guide.md)** - Dependency updates and Dependabot vulnerability management
 - **[Troubleshooting Guide](docs/development/troubleshooting.md)** - TypeScript and build error resolution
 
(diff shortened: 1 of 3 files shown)


prog893:
@coderabbitai review

prog893:
@coderabbitai review

prog893:
@coderabbitai resume

prog893:
@coderabbitai full review

prog893:
@coderabbitai All five addressed in `8ca3bd4`. Two of them changed my mind about what the document should say, so details below.

**Markdown lint.** Five bare fences in the guide tagged `text`, blank lines added around the CLAUDE.md `bash` fence. The four remaining untagged fences in CLAUDE.md (directory trees at L107/L153/L279 and the commit template at L404) are pre-existing and outside this diff, so left alone.

**Rate-limit checker, fail closed.** Correct, and worse than described: folding the fetch into the pipeline meant a `gh` failure reported "still waiting", so a caller sleeping on that would wait forever on an expired token. The fetch is now separate with its own exit code, and only the newest notice is used. Extracted the block verbatim and ran it:

| case | exit |
|---|---|
| window open | 0 |
| still waiting | 1 |
| no notice on the PR | 2 |
| 404 / missing token / `gh` not on PATH | 3 |

Only 0 and 2 mean it is safe to kick.

**"Stayed put" wording.** You are right, and this is the one worth calling out, because the document is about rigor and that sentence was not. The three readings resolve to `01:59:40`, `01:59:55` and `01:59:02`, a spread of 53 seconds. What the data supports is that a kick recomputes the countdown **without restarting the roughly 40 minute quota window**, not that the time is invariant. Rewritten to say exactly that, with the readings shown as a table and the resolution stated as "good to about a minute".

**Silence is ambiguous.** Agreed, and this bit during this very PR: GitHub was returning scattered 503s, which is indistinguishable from quota if you only look at CodeRabbit. The table now marks it ambiguous and there are diagnostics (`gh auth status`, a repo reachability call, an installation check) to run before waiting an hour on a guess.

**Trial-merge cleanup.** Correct: chained with `&&`, cleanup was skipped on exactly the runs where validation failed. Moved to an `EXIT` trap. Verified in a throwaway repo with a deliberately failing step: back on `main`, no trial branch, no `MERGE_HEAD`, no merge residue. My first check appeared to leave a modified file, which turned out to be the test script itself sitting untracked, not merge leftovers.

One addition prompted by getting this PR reviewed at all: the three kick commands are not interchangeable. `review` is incremental and skips commits it has seen, `resume` only un-pauses automatic reviews, and `full review` forces a re-examination. A rate-limited first attempt marks the commits as seen, so afterwards `review` replies "does not re-review already reviewed commits" and `resume` reports success while changing nothing. `full review` is what produced this review. That sequence is now a table in the guide.

prog893:
@coderabbitai Addressed in `264db59`. Valid finding, and it was a defect in the instruction rather than in what was actually done.

The round-trip steps said to capture "the full prior state" and diff it against a fresh read. Read literally that includes the API envelope, so the diff would fail on runs where the restore was perfectly correct. That is worse than a flaky check: it teaches you to ignore the one step that proves the restore worked.

The `#105` round trip happened to project a field subset, so it did not hit this, but the document as written would have led someone straight into it.

Measured before writing it down rather than asserting it:

- **`etag`** is byte-stable across repeated reads of an untouched resource. Two consecutive `videos.list` calls on the same video returned `LFMoZ0QA4OEz5mNKMzZ2Q8PchRM` both times. So a change in `etag` reflects a write, not read noise, and including it in the comparison turns every write test into a false positive.
- **Caption `snippet.lastUpdated`** is write-stamped, visible directly in the archive: a track carrying `2023-11-02T13:35:10Z` sits beside one stamped at the moment of upload.

Step 1 now says to capture a **canonical projection**: every field being changed, every field that might be disturbed, nothing server-managed. The guide shows the actual probe used for the `#105` round trip.

Added one point your finding implies but does not say outright: the projection has to be **wider** than the field under test. `defaultAudioLanguage` was not being changed deliberately, and it was included only out of caution. That caution is the entire reason its drift from `en` to `en-US` was caught rather than shipped. A projection narrowed to "fields under test" alone would have missed it, so the rule is "changing, plus might-be-disturbed, minus server-managed".

prog893:
@coderabbitai Requesting an explicit sign-off review on this PR.

All six findings across both passes are resolved, and you have confirmed the last one yourself:

1. **Markdown fences** (CLAUDE.md, guide) - fixed in `8ca3bd4`. Marked addressed.
2. **Rate-limit checker fail closed, newest notice only** - fixed in `8ca3bd4`. Marked addressed. Exit codes verified by extracting the block verbatim: 0 open, 1 waiting, 2 no notice, 3 lookup failed.
3. **Imprecise "stayed put" wording** - fixed. Marked addressed. Now states the 53-second spread across three readings and claims only that a kick does not restart the quota window.
4. **Missing acknowledgement is ambiguous** - fixed in `8ca3bd4`. Marked addressed. Diagnostics added ahead of any wait.
5. **Trial-merge cleanup on failure** - fixed in `8ca3bd4`. Marked addressed. Moved to an EXIT trap, verified in a throwaway repo with a deliberately failing step.
6. **Canonical projection instead of the raw envelope** - fixed in `264db59`. You replied "Verified in 264db59 ... This addresses the finding."

Nothing is outstanding, but the PR-level verdict is still CHANGES_REQUESTED from the first pass, which no longer matches any open finding. Your later responses came through as comments, which do not supersede it.

Please post an updated review so the verdict reflects the current state. If anything above is still open from your side, say which and I will fix it rather than merge past it.

CI on the head commit: `checks` SUCCESS, `claude-review` SUCCESS. Docs only, no code or behavior change.

Issue #173, opened 2026-08-20
get-video-analytics: default metric set makes most --dimensions unusable


prog893:
## Problem

`get-video-analytics` sends eight metrics by default:

```text
views, estimatedMinutesWatched, averageViewDuration, averageViewPercentage,
likes, dislikes, comments, shares
```

The four engagement metrics are only available for a handful of dimensions, and
the Analytics API rejects the whole query when any requested metric is
unavailable for the requested dimension. So most valid dimensions fail out of
the box with the generic `The query is not supported`, and nothing in the error
points at the metrics.

Measured live on 2026-08-20, one metric at a time, video-level query:

| dimension | views | estimatedMinutesWatched | averageViewDuration | averageViewPercentage | likes | dislikes | comments | shares |
|---|---|---|---|---|---|---|---|---|
| `day` | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| `subscribedStatus` | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ❌ | ✅ |
| `deviceType` | ✅ | ✅ | ✅ | ✅ | ❌ | ❌ | ❌ | ❌ |
| `youtubeProduct` | ✅ | ✅ | ✅ | ✅ | ❌ | ❌ | ❌ | ❌ |

Dropping the single offending metric is enough to fix it: with
`--dimensions subscribedStatus`, the full default set minus `comments` returns
2 rows, while the full set is rejected.

## Impact

Every dimension except `video`, `day` and `creatorContentType` fails on the
default invocation. Measured with default metrics vs `--metrics views`:

```text
--dimensions deviceType                      REJECT   vs  OK  (4 rows)
--dimensions operatingSystem                 REJECT   vs  OK (10 rows)
--dimensions subscribedStatus                REJECT   vs  OK  (2 rows)
--dimensions youtubeProduct                  REJECT   vs  OK  (1 row)
--dimensions liveOrOnDemand                  REJECT   vs  OK  (2 rows)
--dimensions insightTrafficSourceType        REJECT   vs  OK  (8 rows)
--dimensions insightPlaybackLocationType     REJECT   vs  OK  (3 rows)
--dimensions deviceType,operatingSystem      REJECT   vs  OK (11 rows)
--dimensions country,subscribedStatus        REJECT   vs  OK  (3 rows)
```

This is why #143 looked like an allowlist problem from the outside. The
[...]

PR #174, opened 2026-08-20
fix(#143): reconcile the --dimensions allowlist with the live API

### LINKED ISSUE #143: Reconcile get-video-analytics --dimensions allowlist with the live-tested compatibility matrix
## Problem

Porting PR #50's `docs/dimension-compatibility.md` (100+ live-tested combinations) surfaced gaps against the allowlist shipped in #120:

- **`youtubeProduct` and `liveOrOnDemand` work** for video-level queries per the tested matrix but are rejected by the client-side allowlist today.
- `country + day` / `country + deviceType` / `country + operatingSystem` **pass the allowlist but the API rejects them** — candidates for `INVALID_DIMENSION_COMBOS` so users get the fast client-side error instead of the opaque API one.
- The matrix also documents the ≤4-dimensions ceiling, which the allowlist doesn't enforce.

## Fix

In `commands/get-video-analytics.ts`: add the two missing dimensions to `VIDEO_DIMENSIONS`, add the three country-pairs (and >4-dimension check) to the combo validation. Re-verify against the live API before shipping — the matrix data dates from #50's testing and the API evolves.

Reference: `docs/dimension-compatibility.md` (ported in the #50-supersession PR).

prog893:
Closes #143.

## Why

The client-side `--dimensions` allowlist exists so a typo costs nothing and a
known-bad request never spends quota. It only pays for itself if it agrees with
the API. It disagreed in both directions, so it refused work the API would do
and waved through work it would not.

Every claim below was measured against the live Analytics API on 2026-08-20
with a video-level query (`filters: video==<id>`), not taken from the previous
matrix, which turned out to be wrong on two points.

## What changed

**Now accepted (the API supports them; the CLI did not):** `youtubeProduct`,
`liveOrOnDemand`, `dma`. These were unreachable through the CLI at any metric
setting.

**Now rejected (the CLI allowed them; the API never did):**

| dimension | why it could never work |
|---|---|
| `province` | needs a `country==US` filter, which the CLI has no flag to send |
| `city` | refused for video queries even with that filter supplied |
| `insightTrafficSourceDetail` | refused even with an `insightTrafficSourceType` filter |
| `insightPlayerLocationType` | not a real identifier; the API answers `Unknown identifier` |

All four previously produced the API's generic "query is not supported". The
error now names the cause.

**Known-bad combinations: 1 to 10.** Six `country` pairings, `day + month`, and
three device pairings. Each was measured as rejected with `--metrics views`,
the most permissive metric set, so no metric choice rescues them. The previous
lone entry (`creatorContentType + insightTrafficSourceDetail`) is dropped: it
named a dimension no longer on the allowlist, so the per-dimension check threw
first and the combo branch was unreachable. A test now asserts that every combo
entry references only allowlisted dimensions, so this cannot silently recur.

## Two things the issue asked for that the measurements refuted

**The 4-dimension ceiling does not exist.** #143 asked for a `>4` check. The
5-dimension example that justified the limit was
`country,creatorContentType,subscribedStatus,youtubeProduct,day`, which contains
`country,day` and already fails at 2 dimensions. Seven dimensions returns 465
rows. Enforcing the ceiling would have broken working queries, so it is not
enforced.

**`country + month` is valid**, not rejected as the old guide guessed. It is
explicitly excluded from the combo table and covered by a test.

## The larger finding: metrics, not dimensions

Compatibility is a property of the dimensions **and** the metrics together. The
default metric set includes `likes`, `dislikes`, `comments` and `shares`, which
the API offers for only a few dimensions:

| dimension | views / watch-time metrics | likes | dislikes | comments | shares |
|---|---|---|---|---|---|
| `day` | ✅ | ✅ | ✅ | ✅ | ✅ |
| `subscribedStatus` | ✅ | ✅ | ✅ | ❌ | ✅ |
| `deviceType` | ✅ | ❌ | ❌ | ❌ | ❌ |
| `youtubeProduct` | ✅ | ❌ | ❌ | ❌ | ❌ |

So `--dimensions deviceType` fails with the default metrics even though
`deviceType` is valid, and has done so independently of this issue. That is a
separate defect, filed as #173, and deliberately not fixed here: narrowing the
default metric set changes what every user gets back by default.

The allowlist deliberately does not model the interaction. It is validated
against the most permissive metric set and answers only "can this dimension
ever work", which is documented at the definition.

## E2E proof

Identical script on `main` (e2f4cba) and on this branch, 30 real invocations.
Read-only analytics queries; no mutating command involved.

Dimensions this PR unblocks, run with `--metrics views`:

```text
                                    before (main)             after
youtubeProduct                      exit=1 CLIENT-REJECT  ->  exit=0  14 rows
liveOrOnDemand                      exit=1 CLIENT-REJECT  ->  exit=0  15 rows
country,youtubeProduct              exit=1 CLIENT-REJECT  ->  exit=0  18 rows
youtubeProduct,liveOrOnDemand       exit=1 CLIENT-REJECT  ->  exit=0  20 rows
7 dimensions at once                exit=1 CLIENT-REJECT  ->  exit=0 465 rows
dma                                 exit=1 CLIENT-REJECT  ->  exit=0  13 rows
```

Opaque API errors replaced by named client-side errors (exit code unchanged at
1, but no quota spent and the message identifies the problem):

```text
country,day / country,deviceType / country,operatingSystem / day,month
province / city / insightTrafficSourceDetail / insightPlayerLocationType
        API-REJECT  ->  CLIENT-REJECT
```

Controls, all identical before and after:

```text
positive:  --dimensions day                       exit=0  52 rows
           --dimensions country                   exit=0  36 rows
           no --dimensions (default)              exit=0  35 rows
           deviceType,operatingSystem + views     exit=0  38 rows
           country,subscribedStatus + views       exit=0  22 rows
negative:  --dimensions bogus                     exit=1  rejected
           --dimensions ageGroup                  exit=1  rejected
           --dimensions ""                        exit=1  "cannot be empty"
```

18 of 30 lines changed; the other 12 are the controls above.

`bun run type-check`, `bun run lint` (0 errors) and `bun test` all green.
Tests 172 to 186: a new `tests/analytics.test.ts` covering the allowlist, every
combo, the absent arity ceiling, and the reachability invariant.

## Docs

`docs/dimension-compatibility.md` rewritten from the measurements, including the
metric matrix, the corrected `month` date rule, and explicit notes where the
previous revision guessed wrong. `docs/commands/analytics.md` had listed the
removed dimensions and used `--dimensions day,country` as its example, which
this PR now rejects; both are corrected.

## Summary by CodeRabbit

- **Documentation**
  - Updated video Analytics dimension guidance to reflect current API support.
  - Added coverage for time-bucket requirements, valid dimension combinations, metric compatibility, validation errors, quotas, dates, and ownership.
  - Clarified that supported queries are not limited to four dimensions.

- **Bug Fixes**
  - Improved validation for unsupported dimensions and conflicting geography, time, and device combinations.
  - Added support for DMA, YouTube product, and live/on-demand dimensions.

### DIFF
--- a/docs/commands/analytics.md
+++ b/docs/commands/analytics.md
@@ -62,14 +62,17 @@ staqan-yt get-video-analytics --video-id dQw4w9WgXcQ --output json
 
 ### Default Metrics
 
-If no `--metrics` specified, fetches:
+If no `--metrics` specified, fetches these eight (`DEFAULT_VIDEO_METRICS` in `lib/analytics.ts`):
 - `views` - Total views
 - `estimatedMinutesWatched` - Total watch time
 - `averageViewDuration` - Average view duration (seconds)
-- `subscribersGained` - Subscribers gained
-- `subscribersLost` - Subscribers lost
+- `averageViewPercentage` - Average share of the video watched
 - `likes` - Total likes
 - `dislikes` - Total dislikes (if available)
+- `comments` - Total comments
+- `shares` - Total shares
+
+The last four are unavailable for most dimensions, so pass `--metrics views` when using `--dimensions`. See [Dimension Compatibility Guide](../dimension-compatibility.md#metrics-decide-more-than-dimensions-do).
 
 ### Available Metrics
 
--- a/docs/commands/analytics.md
+++ b/docs/commands/analytics.md
@@ -97,15 +97,21 @@ By default, queries aggregate by `video` (the only dimension when `--dimensions`
 
 Supported dimensions for video-level queries:
 
-- `day`, `month` — Time buckets
-- `insightTrafficSourceType`, `insightTrafficSourceDetail` — Traffic sources (aggregate and per-referrer)
-- `creatorContentType` — `SHORT`, `LONG_FORM`, `LIVE`
-- `country`, `province`, `city` — Geography
-- `deviceType`, `operatingSystem` — Device
-- `insightPlaybackLocationType`, `insightPlayerLocationType` — Where playback happened
-- `subscribedStatus` — Subscribed vs non-subscribed viewers
+- `day`, `month` - Time buckets (`month` needs first-of-month dates at both ends)
+- `insightTrafficSourceType` - Traffic sources
+- `creatorContentType` - `SHORT`, `LONG_FORM`, `LIVE`
+- `liveOrOnDemand` - Live vs on-demand playback
+- `country`, `dma` - Geography
+- `deviceType`, `operatingSystem` - Device
+- `youtubeProduct` - Core YouTube, Music, Kids, Gaming
+- `insightPlaybackLocationType` - Where playback happened
+- `subscribedStatus` - Subscribed vs non-subscribed viewers
 
-Combine multiple dimensions with commas (e.g. `--dimensions day,country`). The CLI rejects unknown dimensions and known-bad combinations (e.g. `creatorContentType + insightTrafficSourceDetail` — see #88, fixed in PR #90) with a clear error before any API call.
+Combine multiple dimensions with commas (e.g. `--dimensions day,deviceType`). The CLI rejects unknown dimensions and known-bad combinations (e.g. `country + day`) with a clear error before any API call.
+
+**Most dimensions need `--metrics` narrowed.** The default metric set includes `likes`, `dislikes`, `comments` and `shares`, which the API only offers for a few dimensions, so `--dimensions deviceType` fails until you pass something like `--metrics views`. This is the most common cause of `The query is not supported`.
+
+**→ See [Dimension Compatibility Guide](../dimension-compatibility.md)** for the measured dimension/metric matrix, the rejected combinations, and the `month` date rule.
 
 ---
 
--- a/docs/dimension-compatibility.md
+++ b/docs/dimension-compatibility.md
@@ -1,319 +1,190 @@
 # YouTube Analytics Dimension Compatibility Guide
 
+Which `--dimensions` values `get-video-analytics` accepts, and why some
+combinations are refused before the request is sent.
+
 ## Overview
 
-This guide documents which dimension combinations work with the YouTube Analytics API v2 for video-level queries. These findings are based on comprehensive testing with both regular videos and Shorts.
+Every claim in this guide was measured against the live Analytics API. Where an
+earlier revision of this document guessed, the guess is called out and
+corrected, because two of them were wrong in ways that cost real queries.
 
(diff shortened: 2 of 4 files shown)


prog893:
@coderabbitai Requesting an explicit sign-off review on this PR. Both findings are addressed in 2d6621d.

1. **`Default Metrics` section out of sync with the runtime** - fixed. The section now lists the eight metrics `DEFAULT_VIDEO_METRICS` actually sends: `subscribersGained` and `subscribersLost` removed, `averageViewPercentage`, `comments` and `shares` added. Worth noting the finding landed on the most load-bearing part of the change: three of the four metrics the section was missing are precisely the ones that make dimensional queries fail, so the stale list would have contradicted the new compatibility guide. A pointer to that guide was added alongside.

2. **Combination matrix incomplete / test only checks reachability** - accepted in part.

   The test is fixed: it now pins the exact ten-entry matrix, so dropping a measured pair or adding an unmeasured one fails the suite. Reachability alone could not catch either.

   The premise that a measured combination was omitted does not hold, and the matrix is unchanged at ten entries. The discrepancy was in the prose, not the code: the description said "1 to 11" while the table has always had ten, and the PR body is corrected. Enumerated:

   ```
   country+day, country+deviceType, country+operatingSystem,
   country+insightTrafficSourceType, country+insightPlaybackLocationType,
   country+dma, day+month, month+deviceType,
   insightTrafficSourceType+deviceType, insightPlaybackLocationType+deviceType
   ```

   Two measured-rejected pairs are deliberately excluded, and adding either would be a regression:

   - `creatorContentType + insightTrafficSourceDetail` (the previous lone entry) is unreachable now that `insightTrafficSourceDetail` is off the allowlist. The per-dimension check throws first, so the combo branch is dead code. The new reachability test enforces this.
   - `country + month` is **accepted** by the API given first-of-month dates. It read as rejected only when probed with a misaligned range, where the failure is the date-alignment error rather than a combination error. A test asserts it stays allowed.

CI is green. Nothing else is outstanding, but the verdict is still CHANGES_REQUESTED from the first pass. Please post an updated review, and if anything is still open from your side, say which and I will fix it rather than merge past it.

prog893:
@coderabbitai full review

The previous response concluded "Approved" and found no remaining issue, but the formal verdict on the PR is still CHANGES_REQUESTED from the first pass, and the CodeRabbit check reports "Review rate limited". Requesting a full re-examination so the review state reflects the conclusion. Both original findings are addressed in 2d6621d and CI is green.

PR #180, opened 2026-08-21
fix(#173): reconcile metrics with dimensions, from a full live sweep

### LINKED ISSUE #173: get-video-analytics: default metric set makes most --dimensions unusable

prog893:
Closes #173.

## Why

The API rejects the **entire query** when any requested metric is unavailable
for the requested dimensions, and reports it with the same opaque `The query is
not supported` it uses for dimension conflicts. The default metric set carries
`likes`, `dislikes`, `comments` and `shares`, which most dimensions do not
offer, so nearly every dimension failed on a default invocation:

```text
--dimensions deviceType                   exit=1  API-REJECT
--dimensions operatingSystem              exit=1  API-REJECT
--dimensions subscribedStatus             exit=1  API-REJECT
--dimensions youtubeProduct               exit=1  API-REJECT
--dimensions liveOrOnDemand               exit=1  API-REJECT
--dimensions insightTrafficSourceType     exit=1  API-REJECT
--dimensions insightPlaybackLocationType  exit=1  API-REJECT
--dimensions dma                          exit=1  API-REJECT
--dimensions deviceType,operatingSystem   exit=1  API-REJECT
--dimensions country,subscribedStatus     exit=1  API-REJECT
```

## What changed

Resolved client-side, so no quota is spent discovering it:

- **Metrics left at their default** are narrowed to what the dimensions permit.
  Dropped metrics are named on the progress line, never on stdout, so piped
  output stays clean.
- **Metrics passed explicitly** are never altered. An incompatible one is an
  error naming the metric and listing the alternatives, because returning data
  other than what was asked for is worse than failing.

## Method: two 1-D sweeps, not a cross product

The naive sweep is every dimension-set by every metric-set, which does not
finish. `scripts/sweep-analytics-compat.ts` measures two independent axes by
holding the other at a value that cannot conflict:

- **Dimension validity** is probed with `views`, which every dimension permits,
  so a rejection is always a dimension conflict.
- **Metric validity** is probed one dimension at a time, so a rejection is
  always a metric conflict.

They compose: a query is valid when every **pair** of its dimensions is valid,
no dimension exceeds its **arity cap**, and its metrics sit inside the
**intersection** of the per-dimension metric sets. 296 probes total.

The laws are verified, not assumed. The final phase predicts higher-order cases
and checks each against the API, and the run **exits non-zero without emitting
tables** if any prediction is contradicted.

## What the sweep found that hand-probing missed

**The pair table was a third complete.** All 66 pairs were probed and 24
rejected, against 10 shipped previously. The additions are the `month` device
and insight conflicts (`month+operatingSystem`, `month+insightTrafficSourceType`,
`month+insightPlaybackLocationType`), the insight cross-conflicts
(`youtubeProduct+insightTrafficSourceType`, `operatingSystem+insightPlaybackLocationType`,
`insightTrafficSourceType+insightPlaybackLocationType` and others), and the six
`dma` conflicts. Those were previously reaching the API and failing there.

**`dma` has an arity cap that no pairwise table can express.** It is valid alone
and in any non-conflicting pair, yet every three-dimension query containing it
is rejected:

```text
dma                     OK     day,dma,subscribedStatus            REJECT
day,dma                 OK     day,dma,creatorContentType          REJECT
dma,subscribedStatus    OK     dma,subscribedStatus,creatorConte.. REJECT
dma,creatorContentType  OK     day,dma,month                       REJECT
```

This was found **by the guard, not by luck**: the first sweep predicted
`day,dma,subscribedStatus` would work from the pairwise law, the API disagreed,
and the run refused to generate tables. A per-dimension arity cap now models it.
There is still no global ceiling; a 7-dimension query still returns 465 rows.

## E2E proof

Identical script on `main` (4f34dee) and this branch, 30 real invocations,
read-only throughout.

**Fixed, 11 queries that could not run at all now return data.** Column counts
confirm the narrowing is correct rather than merely non-failing:

```text
                                  before            after
--dimensions deviceType           exit=1 API-REJECT -> exit=0  5 cols
--dimensions operatingSystem      exit=1 API-REJECT -> exit=0  5 cols
--dimensions subscribedStatus     exit=1 API-REJECT -> exit=0  8 cols
--dimensions youtubeProduct       exit=1 API-REJECT -> exit=0  5 cols
--dimensions liveOrOnDemand       exit=1 API-REJECT -> exit=0  4 cols
--dimensions insightTrafficSou..  exit=1 API-REJECT -> exit=0  5 cols
--dimensions insightPlaybackLo..  exit=1 API-REJECT -> exit=0  5 cols
--dimensions dma                  exit=1 API-REJECT -> exit=0  5 cols
--dimensions deviceType,operat..  exit=1 API-REJECT -> exit=0  6 cols
--dimensions country,subscribe..  exit=1 API-REJECT -> exit=0  9 cols
--dimensions dma,day              exit=1 API-REJECT -> exit=0  6 cols
```

`liveOrOnDemand` returning 4 columns (1 dimension + 3 metrics) rather than 5 is
the check working: it is the one dimension that also rejects
`averageViewPercentage`.

**Opaque API errors became named client-side errors**, same exit code, no quota:

```text
explicit views,likes + deviceType     API-REJECT -> CLIENT-METRIC-ERR
month,operatingSystem                 API-REJECT -> CLIENT-DIM-ERR
youtubeProduct,insightTrafficSource.. API-REJECT -> CLIENT-DIM-ERR
dma,youtubeProduct                    API-REJECT -> CLIENT-DIM-ERR
operatingSystem,insightTrafficSour..  API-REJECT -> CLIENT-DIM-ERR
day,dma,subscribedStatus              API-REJECT -> CLIENT-DIM-ERR  (arity cap)
```

**Controls, identical before and after:**

```text
positive:  --dimensions day / country / creatorContentType   exit=0  9 cols
           no --dimensions (default)                         exit=0  9 cols
           explicit --metrics views + deviceType             exit=0  2 cols
           explicit --metrics likes + day                    exit=0  2 cols
negative:  --dimensions bogus / province / day,month         exit=1  rejected
```

The four 9-column rows prove no narrowing happens where none is needed, and the
explicit-metrics rows prove an explicit request is passed through untouched.

`bun run type-check`, `bun run lint` (0 errors) and `bun test` green.
Tests 187 to 197.

## Maintenance

`docs/analytics-compat-snapshot.json` holds the raw sweep output, and
`tests/analytics.test.ts` pins the shipped tables, so drift fails CI instead of
going unnoticed. That is the actual fix for what caused #143: the old matrix
rotted because nothing ever re-checked it. Re-running the sweep is now the
documented maintenance procedure.

## Summary by CodeRabbit

- **New Features**
  - Analytics queries now validate dimension combinations and enforce dimension-specific limits.
  - Compatible metrics are automatically selected for default queries.
  - Explicitly unsupported metrics now produce clear errors, while unknown metrics remain supported.

- **Documentation**
  - Added guidance on metric compatibility, rejected dimension combinations, query limits, and validation behavior.
  - Documented the analytics compatibility scanning workflow and usage.

- **Tests**
  - Expanded coverage for metric filtering, dimension intersections, rejected combinations, and arity limits.

### DIFF
--- a/docs/analytics-compat-snapshot.json
+++ b/docs/analytics-compat-snapshot.json
@@ -1,0 +1,341 @@
+{
+  "generatedAt": "2026-08-21T01:38:25.393Z",
+  "videoId": "eQICyHf3hsY",
+  "dateRange": {
+    "startDate": "2026-06-01",
+    "endDate": "2026-08-01"
+  },
+  "validDimensions": [
+    "video",
+    "day",
+    "month",
+    "country",
+    "dma",
+    "deviceType",
+    "operatingSystem",
+    "subscribedStatus",
+    "youtubeProduct",
+    "liveOrOnDemand",
+    "creatorContentType",
+    "insightTrafficSourceType",
+    "insightPlaybackLocationType"
+  ],
+  "rejectedDimensions": [
+    {
+      "dimension": "province",
+      "verdict": "REJECT"
+    },
+    {
+      "dimension": "city",
+      "verdict": "REJECT"
+    },
+    {
+      "dimension": "continent",
+      "verdict": "REJECT"
+    },
+    {
+      "dimension": "subContinent",
+      "verdict": "REJECT"
+    },
+    {
+      "dimension": "insightTrafficSourceDetail",
+      "verdict": "REJECT"
+    },
+    {
+      "dimension": "insightPlaybackLocationDetail",
+      "verdict": "REJECT"
+    },
+    {
+      "dimension": "insightPlayerLocationType",
+      "verdict": "UNKNOWN_ID"
+    },
+    {
+      "dimension": "ageGroup",
+      "verdict": "REJECT"
+    },
+    {
+      "dimension": "gender",
+      "verdict": "REJECT"
+    },
+    {
+      "dimension": "sharingService",
+      "verdict": "REJECT"
+    },
+    {
+      "dimension": "channel",
+      "verdict": "REJECT"
+    },
+    {
+      "dimension": "playlist",
+      "verdict": "REJECT"
+    },
+    {
+      "dimension": "group",
+      "verdict": "REJECT"
+    },
+    {
+      "dimension": "uploaderType",
+      "verdict": "REJECT"
+    },
+    {
+      "dimension": "elapsedVideoTimeRatio",
+      "verdict": "REJECT"
+    },
+    {
+      "dimension": "audienceType",
+      "verdict": "REJECT"
+    }
+  ],
+  "invalidPairs": [
+    [
+      "day",
+      "month"
+    ],
+    [
+      "day",
+      "country"
+    ],
+    [
+      "month",
+      "deviceType"
+    ],
+    [
+      "month",
+      "operatingSystem"
+    ],
+    [
+      "month",
+      "insightTrafficSourceType"
+    ],
+    [
+      "month",
+      "insightPlaybackLocationType"
+    ],
+    [
+      "country",
+      "dma"
+    ],
+    [
+      "country",
+      "deviceType"
+    ],
+    [
+      "country",
+      "operatingSystem"
+    ],
+    [
+      "country",
+      "insightTrafficSourceType"
+    ],
+    [
+      "country",
+      "insightPlaybackLocationType"
+    ],
+    [
+      "dma",
+      "deviceType"
+    ],
+    [
+      "dma",
+      "operatingSystem"
+    ],
+    [
+      "dma",
+      "youtubeProduct"
+    ],
+    [
+      "dma",
+      "liveOrOnDemand"
+    ],
+    [
+      "dma",
+      "insightTrafficSourceType"
+    ],
+    [
+      "dma",
+      "insightPlaybackLocationType"
+    ],
+    [
+      "deviceType",
+      "insightTrafficSourceType"
+    ],
+    [
+      "deviceType",
+      "insightPlaybackLocationType"
+    ],
+    [
+      "operatingSystem",
+      "insightTrafficSourceType"
+    ],
+    [
+      "operatingSystem",
+      "insightPlaybackLocationType"
+    ],
+    [
+      "youtubeProduct",
+      "insightTrafficSourceType"
+    ],
+    [
+      "youtubeProduct",
+      "insightPlaybackLocationType"
+    ],
+    [
+      "insightTrafficSourceType",
+      "insightPlaybackLocationType"
+    ]
+  ],
+  "arityCaps": {
+    "dma": 2
+  },
+  "dimensionMetrics": {
+    "video": [
+      "views",
+      "estimatedMinutesWatched",
+      "averageViewDuration",
+      "averageViewPercentage",
+      "likes",
+      "dislikes",
+      "comments",
+      "shares",
+      "subscribersGained",
+      "subscribersLost",
+      "videosAddedToPlaylists",
+      "redViews",
+      "estimatedRedMinutesWatched",
+      "annotationClickThroughRate",
+      "cardClickRate"
+    ],
+    "day": [
+      "views",
+      "estimatedMinutesWatched",
+      "averageViewDuration",
+      "averageViewPercentage",
+      "likes",
+      "dislikes",
+      "comments",
+      "shares",
+      "subscribersGained",
+      "subscribersLost",
+      "videosAddedToPlaylists",
+      "redViews",
+      "estimatedRedMinutesWatched",
+      "annotationClickThroughRate",
+      "cardClickRate"
+    ],
+    "month": [
+      "views",
+      "estimatedMinutesWatched",
+      "averageViewDuration",
+      "averageViewPercentage",
+      "likes",
+      "dislikes",
+      "comments",
+      "shares",
+      "subscribersGained",
+      "subscribersLost",
+      "videosAddedToPlaylists",
+      "redViews",
+      "estimatedRedMinutesWatched",
+      "annotationClickThroughRate",
+      "cardClickRate"
+    ],
+    "country": [
+      "views",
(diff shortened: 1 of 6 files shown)


prog893:
@coderabbitai Requesting an explicit sign-off review. All six findings are addressed in 37f7e39, none disputed.

1. **Transient failures recorded as incompatibility verdicts** (major) - fixed, and this was the most valuable finding in the review. The sweep's entire purpose is producing tables trusted enough to ship into `lib/analytics.ts`, so a 429 misread as `REJECT` would have become a permanent client-side rejection of a query the API serves, with nothing downstream able to detect it. Transient responses (429, 5xx, and connection-level failures) are now retried with exponential backoff, and a probe that cannot be resolved aborts the run via `UnresolvedProbeError`. An unrecognized error no longer degrades to `OTHER` either, since that verdict was itself treated as an incompatibility.

2. **Dropped-metric notice invisible** - fixed. Confirmed the mechanism: `onProgress` assigns `spinner.text`, and the chunk loop reassigns it on the next line before a frame renders. Moved to stderr, which also keeps stdout pipeable. Verified live:

   ```
   $ ... --dimensions deviceType --output json 2>/dev/null | python3 -c "import json,sys; json.load(sys.stdin)"
   valid JSON, columns: deviceType, views, estimatedMinutesWatched, averageViewDuration, averageViewPercentage

   $ ... --dimensions deviceType --output json 2>&1 >/dev/null
   Note: dropped likes, dislikes, comments, shares from the default metrics; not available for --dimensions deviceType.
   ```

3. **Snapshot written before the law check** - fixed. `writeFileSync` and its log now run only after the `lawChecks.failed` gate, and the contradictions are printed on stderr. The docs and `scripts/README.md` wording was already the stronger claim, so the code now matches it rather than the wording being weakened.

4. **"bisected" is inaccurate** - fixed. Phase C probes the whole set and falls back to probing each metric individually. Corrected in the header comment, the log line, and `scripts/README.md`.

5. **Pair count not reproducible** - fixed. The guide now states that `video` is excluded from pairing, being a filter-style passthrough rather than a breakdown, which is what makes 13 supported dimensions yield 12 pairable ones and 66 pairs.

6. **MD031 blank line before fence** - fixed in the new section.

Re-ran the full 30-command E2E capture after these changes: byte-identical to the pre-fix run, so none of the six altered behavior beyond what was intended. `bun run type-check`, `bun run lint` (0 errors) and `bun test` (197) all green.

Nothing is outstanding, but the verdict is still CHANGES_REQUESTED from the first pass. Please post an updated review, and if anything is still open, say which and I will fix it rather than merge past it.

Issue #178, opened 2026-08-20
Warn when an analytics or report date range spans the 2026-08-24 view-counting change


prog893:
## Context

Every query the CLI issues over a date range that straddles **2026-08-24** returns a series whose `views` column is measured two different ways: old logic before the cutoff, new logic from it onward. This affects both API surfaces:

- `fetchVideoAnalytics` (`lib/analytics.ts:167-200`) splits the range into 90-day chunks and concatenates the rows. With the default `dimensions: 'video'` each chunk contributes one aggregate row, so a caller summing them adds pre- and post-change totals together.
- `fetchChannelAnalytics` (`lib/analytics.ts:545-552`) sends the raw range in one query.
- `fetchReportData` (`lib/reports.ts`) merges daily report rows across the boundary.

The default date range makes this the common case, not an edge case: `fetchVideoAnalytics` defaults `startDate` to the video's upload date, so every all-time query on an existing video will span the cutoff from 2026-08-24 onward.

## Proposed work

1. Add a single exported constant, for example `VIEW_COUNTING_CHANGE_DATE = '2026-08-24'`, in `lib/analytics.ts` or `lib/utils.ts`, with a comment linking the announcement.
2. When a resolved range starts before that date and ends on or after it, warn once on stderr through the existing `progress()` path, naming the metric affected and pointing at `engagedViews` for a consistent series.
3. Include the same signal in the structured results so the MCP server does not have to re-derive it.
4. Cover it in `docs/commands/analytics.md` and `docs/commands/reporting-api.md`, and in `docs/troubleshooting.md` under the heading someone will actually search for, which is a sudden jump in views.

Warning only, not filtering or splitting. The data is still correct for what it measures, and silently reshaping a range would be worse than reporting the discontinuity.

## References

- [How YouTube counts views](https://support.google.com/youtube/answer/2991785)
- [Analytics API metrics reference](https://developers.google.com/youtube/analytics/metrics)

Issue #175, opened 2026-08-20
Analytics: add engagedViews support so pre-2026-08-24 view semantics stay queryable


prog893:
## Context

YouTube is aligning view counting across all formats on **2026-08-24**: a view is counted from the first frame of playback, with no minimum watch time. The stricter pre-change definition survives as **Engaged Views**, exposed as `engagedViews` in Analytics Targeted Queries and `engaged_views` in Bulk Reports. Engaged views, not views, remain the metric tied to monetization and YPP eligibility, so it is the number that matters for most performance analysis, and it is also the only way to compare post-change data against anything archived before the cutoff.

The CLI currently has no notion of this metric anywhere.

## Where `views` is hardcoded today

- `lib/analytics.ts:128` `DEFAULT_VIDEO_METRICS` (eight metrics, `views` first)
- `lib/analytics.ts:288` `fetchTrafficSources` sends `metrics: 'views'`
- `lib/analytics.ts:325` `fetchSearchTerms` sends `metrics: 'views'`
- `lib/analytics.ts:402-418` `CHANNEL_REPORT_TYPES` predefined reports, all `views,estimatedMinutesWatched`
- `lib/analytics.ts:674` `SEARCH_TERMS_METRICS`
- `lib/analytics.ts:751` channel search terms query
- `commands/mcp.ts:245,263` MCP tool description and default metric list

## Proposed work

1. After 2026-08-24, verify against the live API which of these accept `engagedViews`, both alone and alongside `views`. The metric is new, so nothing here should be assumed.
2. Decide whether `engagedViews` joins `DEFAULT_VIDEO_METRICS`. Note the constraint recorded in #173: the default set already fails for most `--dimensions` values, so widening it has a cost. An explicit opt-in flag may be better than a wider default.
3. Extend `docs/commands/analytics.md` metric lists (lines 66-88, 214, 422) with `engagedViews` and a one-line statement of the difference between the two.
4. Re-run the `docs/dimension-compatibility.md` matrix with `engagedViews` as the probe metric. The matrix was measured with `--metrics views` and its conclusions do not automatically carry over.
5. Mirror the new metric in the MCP tool schema so agent callers can request it.

## References

- [Data API revision history](https://developers.google.com/youtube/v3/revision_history)
- [Analytics API revision history](https://developers.google.com/youtube/analytics/revision_history)
- [Analytics API metrics reference](https://developers.google.com/youtube/analytics/metrics)
[...]

Issue #70, opened 2026-06-06
Bug: get-channel-analytics does not error when --report and --dimensions/--metrics are used together


prog893:
## Summary

Using `--report` together with `--dimensions` and `--metrics` produces no conflict error. The predefined report template silently wins and the custom dimensions/metrics are ignored, giving the user no feedback.

## Repro

```bash
staqan-yt get-channel-analytics --report demographics --dimensions deviceType --metrics views
# Runs with --report only; --dimensions and --metrics are silently ignored
```

## Root cause

`commands/get-channel-analytics.ts` does not check for mutual exclusion between `--report` (predefined template) and `--dimensions`/`--metrics` (custom query mode). These are two separate usage modes.

## Fix

```ts
if (options.report && (options.dimensions || options.metrics)) {
  error('Cannot combine --report with --dimensions/--metrics. Use one mode at a time.');
  process.exit(1);
}
```

prog893:
Fixed in #85/#86 (merged 2026-06-29): `get-channel-analytics` now explicitly rejects `--report` combined with `--dimensions`/`--metrics` (checked via presence, not truthiness, so it also catches `--dimensions ""`). Closing — not auto-linked at merge time.

Issue #165, opened 2026-08-11
Dead code: bash completion's `--privacy` arm under `case "$prev"` is unreachable


prog893:
## Summary

In the generated bash completion, the `--privacy)` arm under `case "${prev}" in` (`lib/completion.ts:296-297`) is unreachable. Every input that could reach it is intercepted earlier by the variadic flag scan.

## Why it is dead

The variadic scan at `lib/completion.ts:239-268` runs first and guards on `[[ "${cur}" != -* ]]`. It walks backwards from `cword-1`, and its own `--privacy)` arm returns.

That splits the input space with no remainder:

- **`cur` is not a flag** (the normal case, e.g. `--privacy <TAB>` or `--privacy pu<TAB>`): the scan finds `--privacy` immediately behind the cursor and returns. Control never reaches the `case "${prev}"` block.
- **`cur` starts with `-`**: the scan is skipped, so `case "${prev}"` is reached, but `compgen -W "public private unlisted" -- "-…"` matches nothing, since no candidate begins with a hyphen.

## Evidence

Driving `_staqa_nyt_completion` directly with the two paths instrumented to emit distinguishable candidates (`SCAN_*` from the variadic scan, `PREV_*` from the `$prev` arm):

| input | COMPREPLY |
|---|---|
| `list-videos --privacy` + `cur=""` | `SCAN_public SCAN_private SCAN_unlisted` |
| `list-videos --privacy` + `cur="-"` | *(empty)* |

`PREV_*` is never produced.

## History

Both the scan and the arm arrived together in 16d73c0 ("feat: add --privacy filter and visibility display to list-videos", 2026-03-19), so the arm has been shadowed since it was introduced. No behavior regression is involved.

## Impact

None at runtime, purely dead code. It is a maintenance trap: it reads as the handler for `--privacy` value completion, so an edit made there (adding a value, changing wording) would appear to do nothing.

## Suggested fix

Remove the arm. The zsh side has the equivalent property, already documented in `lib/completion.ts` as of #164: the pre-check owns every value of a variadic flag and the `_arguments` action only completes the flag name. Unlike zsh, the bash arm has no residual role, since flag-name completion there comes from the per-command `compgen -W` lists.

Worth confirming the same reasoning for `--video-ids` / `--playlist-ids`, which have arms in both the scan (`:258-261`) and the `case "${prev}"` block (`:274-279`) and may be dead for the same reason.

## Related

[...]

PR #181, opened 2026-08-21
fix(#179): rank custom analytics queries by any selected view metric

### LINKED ISSUE #179: get-channel-analytics: custom-query sort is hardcoded to views, so an engagedViews-only query returns unsorted rows
## Context

`fetchChannelAnalytics` picks the sort field for custom queries by looking for the literal string `views` in the requested metrics (`lib/analytics.ts:510-513`):

```ts
// The API rejects a sort field that isn't in metrics/dimensions, so only
// sort custom queries that actually select views.
sort = params.metrics.split(',').map(m => m.trim()).includes('views') ? '-views' : undefined;
```

Once `engagedViews` becomes available on 2026-08-24, this has two failure modes.

**Unsorted output.** `--metrics engagedViews --dimensions country` selects a view metric, but the exact-match check does not fire, `sort` stays `undefined`, and the API returns rows in its own order. A top-countries query silently stops being ranked. The guard is correct in refusing to send `-views` here, since the API rejects a sort field absent from the metric list, but falling through to no sort at all is not the right answer.

**Wrong ranking axis.** `--metrics views,engagedViews` sorts by `-views`, the new inclusive count, when the caller asking for both almost certainly wants the engaged figure ranked, or at least wants to say which.

The predefined report path is unaffected: `CHANNEL_REPORT_TYPES` carries an explicit `sort` per report (`lib/analytics.ts:402-418`).

## Proposed work

1. Replace the literal check with a preference order over the requested metrics, for example first match among `views`, `engagedViews`, `estimatedMinutesWatched`, so any sortable metric in the list is used rather than only `views`.
2. Consider a `--sort` flag for custom queries so the axis is explicit instead of inferred. It also removes the guesswork for metric sets that contain neither view metric.
3. Add `engagedViews` to the predefined `CHANNEL_REPORT_TYPES` metric sets, or leave them alone deliberately and say why in the comment. Either is defensible, silence is not.
4. Test coverage for the sort selection, which is currently untestable without a live call.

## References

- [Analytics API reports.query sort parameter](https://developers.google.com/youtube/analytics/reference/reports/query)

prog893:
Closes #179.

## The problem

`fetchChannelAnalytics` picked the sort field for a custom query by testing the requested metrics for the literal string `views`:

```ts
sort = params.metrics.split(',').map(m => m.trim()).includes('views') ? '-views' : undefined;
```

Any metric set that did not contain exactly `views` fell through to no sort at all, and the API's own row order stood. A top-N query silently stopped being a ranking.

This is not a future problem. `engagedViews` is **already accepted by the live API** ahead of the 2026-08-24 view-counting change, so the failure reproduces today.

## Live measurement (2026-08-21, `--dimensions day`, range 2026-08-05..2026-08-19)

| `--metrics` | row order returned |
|---|---|
| `views` | ranked: 56, 55, 54, 34, 32, ... |
| `engagedViews` | **raw date order**: 08-05, 08-06, 08-07, ... |
| `estimatedMinutesWatched` | **raw date order** |

`views` and `engagedViews` already diverge in historical data, well before the cutoff. Country `JP` over 2026-05-01..2026-08-19: **1404 views, 802 engaged views**. Engaged views is the figure that keeps the pre-change semantics and the one tied to monetization, so the metric most worth ranking was the one losing its ranking.

## The change

**Preference order instead of a literal check.** A custom query is ranked by the first of `views`, `engagedViews`, `estimatedMinutesWatched` present in the selected metrics, descending. `views` stays first, so no query that sorted before this change moves. Extracted into a pure exported `resolveCustomSort`, which makes the selection testable without a live call (issue item 4).

**A `--sort` flag**, because the preference order cannot resolve intent when both view metrics are selected: `--metrics views,engagedViews` ranks by `-views`, and a caller who wants the engaged figure ranked previously had no way to say so. It accepts any dimension or metric the query selects, `-field` for descending, and is validated client-side so an unsortable axis is named here rather than surfacing as the API's opaque 400.

**`--report` + `--sort` is rejected**, matching the existing #70 check. A predefined report fixes its own sort, so honoring the flag would silently redefine the report and ignoring it would silently discard the flag.

**Predefined report metric sets deliberately keep no `engagedViews`** (issue item 3). Each is a fixed output shape parsed by column, and whether `engagedViews` is accepted for each of those dimensions has not been swept the way #173 swept the rest, so adding it would ship an unverified claim. The reasoning is recorded at `CHANNEL_REPORT_TYPES` and the work left to #175.

## E2E proof

Baseline captured from the released binary before the change; identical commands re-run against `bun dist/bin/staqan-yt.js` after.

**Regression control.** `--metrics views` over the same range is **byte-identical** to main (`diff` clean). The metric that already sorted did not move.

**Fix.** `engagedViews` and `estimatedMinutesWatched` now come back ranked descending, and the row *set* is unchanged, only reordered (`diff` of both sorted, clean) so nothing was dropped or filtered.

**Report path.** `--report geography` output unchanged.

**Negative controls**, all exit 1 with a message naming the problem:

| command | message |
|---|---|
| `--dimensions country --metrics views --sort likes` | `Invalid --sort value: "likes". ... Available: country, views.` |
| `--report geography --sort -views` | `Cannot combine --report with --sort. ...` |
| `--dimensions country --metrics views --sort -` | `--sort cannot be empty` |
| `--report geography --metrics views` | `Cannot combine --report with --dimensions or --metrics.` (#70, still rejected) |

**Positive controls.** `--sort day` returns chronological order; `--sort -likes` overrides the `views` default and is accepted by the API.

Tests 197 -> 206. `type-check`, `lint`, `build`, `bun test` all green.

## Note for #178

The measurement above weakens a premise in #178, which proposes warning that a range spanning 2026-08-24 mixes old- and new-logic views. In Analytics, `views` and `engagedViews` already differ across all of July and early August 2026, so there may be no clean discontinuity at that date on this API surface. Worth confirming against post-cutoff data before shipping a warning that asserts one.

## Summary by CodeRabbit

* **New Features**
  * Added sorting for custom channel analytics queries.
  * Supports ascending and descending sorting using the selected dimensions and metrics.
  * Analytics results now use a sensible default metric order when no sort field is specified.
  * Added sorting support to the analytics tool and shell completion.

* **Bug Fixes**
  * Added validation for unsupported sort fields and incompatible predefined reports.

* **Documentation**
  * Documented sorting syntax, defaults, validation, and limitations.

### DIFF
--- a/bin/staqan-yt.ts
+++ b/bin/staqan-yt.ts
@@ -488,6 +488,7 @@ program
   .option('--end-date <date>', 'End date (YYYY-MM-DD), defaults to today')
   .option('--dimensions <dims>', 'Custom dimensions (comma-separated, requires --metrics)')
   .option('--metrics <metrics>', 'Custom metrics (comma-separated, requires --dimensions)')
+  .option('--sort <field>', 'Sort custom query by a selected field, "-field" for descending (not valid with --report)')
   .option('--output <format>', 'Output format: json, table, text, pretty, csv')
   .option('-v, --verbose', 'Enable verbose output with debug information')
   .action(withHelpWrapper('get-channel-analytics', getChannelAnalyticsCommand));
--- a/commands/get-channel-analytics.ts
+++ b/commands/get-channel-analytics.ts
@@ -31,6 +31,7 @@ async function getChannelAnalyticsCommand(options: ChannelAnalyticsOptions): Pro
         report: options.report,
         dimensions: options.dimensions,
         metrics: options.metrics,
+        sort: options.sort,
         onProgress: (message) => { spinner.text = message; },
       });
     } catch (analyticsErr) {
--- a/commands/mcp.ts
+++ b/commands/mcp.ts
@@ -228,6 +228,10 @@ const TOOLS: Tool[] = [
           type: 'string',
           description: 'Custom metrics (comma-separated, requires dimensions)',
         },
+        sort: {
+          type: 'string',
+          description: 'Sort a custom query by one of its selected dimensions or metrics, "-field" for descending. Not valid with report. Defaults to the first of views, engagedViews, estimatedMinutesWatched present in metrics; a query selecting none of those is returned in the API\'s own order unless sort is given.',
+        },
         startDate: {
           type: 'string',
           description: 'Start date in YYYY-MM-DD format (defaults to 30 days ago)',
@@ -670,6 +674,7 @@ async function handleToolCall(name: string, args: any) {
         report: args.report,
         dimensions: args.dimensions,
         metrics: args.metrics,
+        sort: args.sort,
       });
 
       if (result.rows.length === 0) {
--- a/docs/commands/analytics.md
+++ b/docs/commands/analytics.md
@@ -299,6 +299,7 @@ staqan-yt get-channel-analytics [channelHandle]
 - `--end-date <date>` - End date (YYYY-MM-DD), defaults to today
 - `--dimensions <dims>` - Custom dimensions (comma-separated, requires `--metrics`)
 - `--metrics <metrics>` - Custom metrics (comma-separated, requires `--dimensions`)
+- `--sort <field>` - Sort a custom query by one of the fields it selects; prefix with `-` for descending. Not valid with `--report`
 - `--output <format>` - Output format: json, table, text, pretty, csv (default: pretty)
 - `-v, --verbose` - Enable verbose output with debug information
 - `-h, --help` - Show help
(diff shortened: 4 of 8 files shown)


prog893:
@coderabbitai full review

The first review attempt was rate limited before it started, so the commits on this PR were marked as seen without ever being examined. Requesting a full review rather than an incremental one for that reason.

prog893:
@coderabbitai Requesting an explicit sign-off review. Both findings were valid and are fixed in `bdd5687`.

**1. `lib/completion.ts:330`, bash `--sort` offering comment-sort values.** Correct, and a regression this PR introduced. The `--sort|-s)` arm under `case "$prev"` was not scoped to a command, so adding `--sort` to `get-channel-analytics` made it offer `top`/`new` there, both of which fail `resolveCustomSort`.

Fixed by scoping on `$cmd`, matching the `--type)` arm directly above it rather than adding a new pattern. Channel analytics sort values stay unenumerated, since a valid field is only knowable from the `--dimensions`/`--metrics` typed so far.

Verified against the generated script in a real bash shell with a stubbed `_init_completion`:

| input | before | after |
|---|---|---|
| `list-comments --sort` | `top new` | `top new` |
| `list-comments -s` | `top new` | `top new` |
| `get-channel-analytics --sort` | `top new` | (empty) |
| `get-channel-analytics --output` | `json table text pretty csv` | unchanged (control) |

While confirming this I also found the zsh `_arguments` block for `get-channel-analytics` never gained `--sort` at all, so the flag did not complete in zsh. Added there with an unenumerated value, matching how `--dimensions` and `--metrics` are specified in the same block.

**2. `docs/commands/analytics.md`, future change described as current.** Correct. The example comment read "the post-2026-08-24 views count", which asserts a change that has not taken effect. Reworded to name the two metrics without dating either, and the surrounding note now states plainly that `engagedViews` is available today and is the metric that keeps the stricter definition once the change lands on 2026-08-24.

Worth recording why this one mattered beyond wording: `views` and `engagedViews` already diverge in historical data (`JP`, 2026-05-01..2026-08-19: 1404 vs 802), so it is genuinely not yet established what the cutoff does to the Analytics `views` series. The docs should not have implied it was settled.

**Out of scope, filed separately as #182.** Testing the fix surfaced a pre-existing bug: per-command arms in the bash completion set `COMPREPLY` and fall through without returning, so the trailing `if [[ "${cur}" == -* ]]` global-options fallback overwrites them. Every command's own flags are unreachable in bash (`get-channel-analytics --s<TAB>` yields nothing; `--<TAB>` yields only `--help --output --verbose`). It predates this PR, affects all 34 commands equally, and is the same family as #165, so it is filed rather than fixed here.

`type-check`, `lint`, `build` and `bun test` (206 pass) are green. Nothing is outstanding from my side, but the verdict is still CHANGES_REQUESTED from the first pass. Please post an updated review, and say if anything is still open rather than leaving it for a merge past a stale verdict.

Issue #177, opened 2026-08-20
Reporting cache: merged bulk reports mix pre/post 2026-08-24 views and a new engaged_views column


prog893:
## Context

Bulk Reports gain an `engaged_views` column when the view-counting change lands on **2026-08-24**, and the existing `views` column changes meaning from that date forward. The report archive under `~/.staqan-yt/data/<channelId>/reports/` holds CSVs downloaded before the cutoff, and `fetchReportData` merges them with freshly downloaded ones into a single result set.

## What breaks

`lib/reports.ts:980-1000` merges cached and fresh rows:

```ts
let allData = [...cachedSurviving, ...fetchedData];
```

Rows are `Record<string, string>` keyed by CSV header, so the merge does not fail loudly. It produces a heterogeneous set instead:

- Pre-cutoff rows have no `engaged_views` key at all.
- Pre-cutoff `views` values are computed under the old logic, post-cutoff ones under the new logic, in the same column of the same output.

The formatters union all keys across rows (`lib/formatters.ts:87-91` for table, `:220` for CSV), so the output gains an `engaged_views` column that is populated for recent dates and empty for older ones, with no explanation. A consumer summing `views` over a range that straddles the cutoff gets a number that means nothing.

`ReportMetadata.columns` is already persisted per cached report (`lib/reports.ts:947`, written through `saveReportToCache`), so the schema difference is detectable without a new fetch.

## Proposed work

1. Compare `columns` across the reports contributing to a merged result. When they differ, emit a warning through the existing `progress()` stderr channel naming the columns present in only some inputs.
2. Decide what the merged rows should carry for a missing `engaged_views`: absent key, or explicit empty string so the CSV shape is stable. Absent is the current accidental behaviour and it makes the two cases indistinguishable from a genuinely empty cell.
3. Surface the schema mismatch in the structured result (alongside `uncoveredRanges`) so the MCP consumer sees the same signal as the CLI rather than inferring it.
4. Consider whether `pickNewestPerWindow` and the cache index need a schema-version field, given #106 already tracks cache index migration.

## References

- [Bulk report channel reports reference](https://developers.google.com/youtube/reporting/v1/reports/channel_reports)
- [Reporting API revision history](https://developers.google.com/youtube/reporting/v1/revision_history)

prog893:
## Measured against the archive: the "new column" premise does not hold

[...]

PR #184, opened 2026-08-24
fix(#178): flag ranges whose views predate the 2026-08-24 counting change

### LINKED ISSUE #178: Warn when an analytics or report date range spans the 2026-08-24 view-counting change

prog893:
Closes #178.

## The problem

On **2026-08-24** YouTube began counting a view from the first frame of playback, with no minimum watch time. The field name did not change, so a `views` column spanning that date carries two definitions with nothing marking the boundary, and a total across it mixes them.

## When the notice fires

Two conditions, both required, so the notice appears exactly when the returned numbers can be misread:

| condition | fires |
|---|---|
| range reaches back before 2026-08-24 | yes |
| range starts on or after 2026-08-24 | **no** |
| selects `views` or `redViews` | yes |
| selects only `engagedViews` | **no**, it keeps the stricter definition on both sides |
| selects neither (`likes,shares`) | **no** |

**Deliberately not limited to ranges that straddle the date.** A range lying entirely before it is internally consistent but is not comparable to later data, which is the same trap one step removed. It gets a notice worded for that case:

```
# straddling
Note: views changes definition inside this date range. 2026-08-20 to 2026-08-23
is counted under the previous definition; 2026-08-24 onward counts every playback
from the first frame. Totals across the whole range mix the two. engagedViews
keeps the stricter definition throughout. https://support.google.com/youtube/answer/2991785

# entirely before
Note: views for 2026-08-01 to 2026-08-19 is counted under the definition used
before 2026-08-24, so it is not comparable to views for dates from 2026-08-24
onward, which counts every playback from the first frame. ...
```

The notice names **the dates affected**, not just the range: a straddling query reports the pre-change slice ending `2026-08-23`, not the range end.

## Where it appears

**stderr, for `--output pretty` and `text` only.** `json`, `csv` and `table` are what scripts read, and a note they cannot parse is noise there. stderr rather than stdout means even the human formats never put it in a redirected file.

**MCP callers get the same signal structurally**, as `viewCountingNotice` on the result (issue item 3), carrying `changeDate`, `affectedMetrics`, `affectedRange` and `spansChange` so the date is not hardcoded downstream. Verified against the live API:

```json
{ "changeDate": "2026-08-24", "affectedMetrics": ["views"],
  "affectedRange": { "startDate": "2026-08-20", "endDate": "2026-08-23" },
  "spansChange": true, "message": "...", "learnMoreUrl": "..." }
```

The CLI's own json payload is **unchanged**, keeping the output shape stable for existing consumers.

## A correction to the issue's premise

#178 proposes warning that a spanning range mixes old- and new-logic views, implying the two metrics agree before the change and diverge after it. **They do not.** Live measurement on 2026-08-21, before the cutoff: channel STAQAN, country `JP`, 2026-05-01..2026-08-19 gave **1404 views against 802 engagedViews** for the same range.

So this notice states only what the date arithmetic supports, which is *which dates were measured under which definition*. It makes no claim about the relationship between the two metrics. The docs say the same explicitly, so the cutoff is not read as a clean split between them.

## E2E proof (live API)

| case | result |
|---|---|
| `--output pretty` / `text`, spanning range | notice on stderr, stdout untouched |
| `--output json` / `csv` / `table`, same range | **silent on both streams** |
| range `2026-08-24..2026-08-24` | silent |
| `--metrics engagedViews`, range reaching back | silent |
| `--metrics likes`, range reaching back | silent |
| range fully before the change | notice, "not comparable" wording |
| `get-video-analytics` all-time (upload 2023-11-02) | notice, `2023-11-02 to 2026-08-23` |

json still parses and exits 0.

## Docs

- `docs/commands/analytics.md`: a section on the change up front, what it means for each case, and the worked example.
- `docs/troubleshooting.md`: filed under **"Views jumped suddenly, or two runs disagree"** (issue item 4 asks for the heading someone will actually search), including that the Data API's `viewCount` has no `engagedViews` equivalent.

## Scope

Analytics surfaces only. The Reporting API's merged-cache problem is a schema mismatch (pre-cutoff CSVs have no `engaged_views` column at all) and belongs to **#177**, which tracks it. `views` in `CHANNEL_REPORT_TYPES` is untouched, per the reasoning recorded in #181.

Tests: 43 in `analytics.test.ts`, covering the 2026-08-23/24 boundary in both directions and metrics that merely contain an affected name (`redViewsPercentage`). `type-check`, `lint`, `build` green.

## Summary by CodeRabbit

* **New Features**
  * Added notices for analytics covering YouTube’s view-counting change on August 24, 2026.
  * Notices identify affected metrics and date ranges, with a link to learn more.
  * Human-readable CLI output displays notices on stderr, keeping structured output unchanged.
  * Analytics results now include view-counting details for integrations such as MCP.

* **Documentation**
  * Added guidance on mixed historical date ranges, affected metrics, and comparing results consistently.
  * Added troubleshooting information and updated analytics tool descriptions.

### DIFF
--- a/commands/get-channel-analytics.ts
+++ b/commands/get-channel-analytics.ts
@@ -71,6 +71,14 @@ async function getChannelAnalyticsCommand(options: ChannelAnalyticsOptions): Pro
     const { channelId, channelTitle, reportType, dateRange, columnHeaders, rows } = result;
     const { startDate, endDate } = dateRange;
 
+    // #178: human-facing formats only. json/csv/table are what scripts read,
+    // and a note they cannot parse is noise there; the same signal reaches
+    // those callers as `viewCountingNotice` in the MCP result. stderr, not
+    // stdout, so even here it never lands in a redirected file.
+    if (result.viewCountingNotice && (outputFormat === 'pretty' || outputFormat === 'text')) {
+      process.stderr.write(`${result.viewCountingNotice.message}\n`);
+    }
+
     if (rows.length === 0) {
       console.log('');
       console.log(chalk.yellow('No analytics data available for this channel and time period.'));
--- a/commands/get-video-analytics.ts
+++ b/commands/get-video-analytics.ts
@@ -41,6 +41,14 @@ async function getVideoAnalyticsCommand(options: AnalyticsOptions): Promise<void
 
     spinner.succeed(`Retrieved ${allRows.length} row(s) of analytics data`);
 
+    // #178: human-facing formats only, on stderr. See get-channel-analytics
+    // for the reasoning. This command is the more exposed of the two, since
+    // the default start date is the upload date, so every all-time query on a
+    // video published before the change reaches back past it.
+    if (result.viewCountingNotice && (outputFormat === 'pretty' || outputFormat === 'text')) {
+      process.stderr.write(`${result.viewCountingNotice.message}\n`);
+    }
+
     // Prepare aggregated data for structured formats
     const aggregated: { [key: string]: number } = {};
     columnHeaders.forEach((header, index) => {
--- a/commands/mcp.ts
+++ b/commands/mcp.ts
@@ -678,11 +678,19 @@ async function handleToolCall(name: string, args: any) {
       });
 
       if (result.rows.length === 0) {
+        // The notice describes the requested date range, not the rows, so it
+        // still applies when nothing came back: a caller comparing this empty
+        // result against pre-cutoff archived data needs the same caveat. The
+        // CLI already emits its notice before its own empty-rows branch, and
+        // dropping it here was the one place the two surfaces disagreed.
+        const noData = 'No analytics data available for this channel and time period.';
         return {
           content: [
             {
               type: 'text',
-              text: 'No analytics data available for this channel and time period.',
+              text: result.viewCountingNotice
+                ? `${noData}\n\n${JSON.stringify({ viewCountingNotice: result.viewCountingNotice }, null, 2)}`
+                : noData,
             },
           ],
         };
--- a/commands/mcp.ts
+++ b/commands/mcp.ts
@@ -207,7 +207,7 @@ const TOOLS: Tool[] = [
   },
   {
     name: 'youtube_get_channel_analytics',
-    description: 'Get channel-level analytics reports from YouTube Analytics API (demographics, devices, geography, traffic sources, subscription status, etc.). When the result carries `viewCountingNotice`, the requested range reaches back before the 2026-08-24 view-counting change, so `views` in these rows is not measured consistently across the range (or against later data); report that caveat rather than presenting a total as comparable. `engagedViews` keeps the stricter pre-change definition throughout.',
+    description: 'Get channel-level analytics reports from YouTube Analytics API (demographics, devices, geography, traffic sources, subscription status, etc.). When the result carries `viewCountingNotice`, the requested range reaches back before the 2026-08-24 view-counting change, so the metrics named in its `affectedMetrics` (`views` and/or `redViews`) are not measured consistently across the range (or against later data); report that caveat rather than presenting a total as comparable. `engagedViews` keeps the stricter pre-change definition throughout.',
     inputSchema: {
       type: 'object',
       properties: {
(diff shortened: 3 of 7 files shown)


prog893:
@coderabbitai full review

prog893:
@coderabbitai Requesting an explicit sign-off review. One finding fixed in `71d2a59`, one declined with measurement.

**1. `commands/mcp.ts:210,249`, describe `redViews` as affected. Correct, fixed.**

Both descriptions told the caller to qualify `views`, so a `redViews`-only query carried a notice pointing at a metric it never selected. Both now point at the result's own `affectedMetrics` rather than naming a metric inline, so the description cannot drift from what the notice reports. The `engagedViews` continuity sentence was also missing from the video-analytics description and is now in both.

Worth noting the runtime message was already correct: it interpolates the matched metrics, so a `redViews`-only query has always produced "Note: redViews changes definition inside this date range...". The defect was confined to the tool descriptions. Added a test asserting a `redViews`-only query names `redViews` and never bare `views`.

I also recorded in the code why `redViews` is on the affected list at all: it counts the same playback event restricted to Premium members, so a change to what counts as a view necessarily moves it. That is an inference from the announcement, not a separately documented statement about `redViews`, and the comment now says so rather than leaving it to read as verified.

**2. `lib/analytics.ts:669`, convert `viewCountingNoticeFor` to a named parameter object. Declining.**

The quoted guideline is about the CLI surface, not TypeScript signatures. In context (`CLAUDE.md:236`, under "AWS API Naming Conventions") every example under it is a shell command:

```
**🚨 RULE: Everything is a named flag. No positional arguments at all.**

get-videos --video-ids <id1> <id2>
```

It governs `commands/` and `bin/staqan-yt.ts`, where this PR adds no arguments at all.

Measured against the codebase, exported functions in `lib/` with two or more parameters:

| shape | count |
|---|---|
| positional | **28** |
| named parameter object | **0** |

Including `cache.ts:saveReportToCache` (5), `cache.ts:findCachedReports` (4), `formatters.ts:formatData` (3). The immediate neighbour is decisive: `resolveCustomSort(dimensions, metrics, explicit)` sits ~40 lines above in this same file, takes three positional parameters in the same shape, and you approved it on `#181` on 2026-08-21.

The codebase does use a params object, but for a different shape: `fetchVideoAnalytics(params: VideoAnalyticsParams)` and friends take 6+ mostly-optional fields configuring an API call. `viewCountingNoticeFor` takes three required strings. Converting only this one would make it the sole object-param helper in `lib/`, so the change would reduce consistency rather than increase it.

Happy to be shown otherwise if there is a guideline that does target function signatures, in which case the right fix is the other 28 in a dedicated PR, not this one.

`type-check`, `lint`, `build` and `bun test` (216 pass) are green. Please post an updated review, and say if anything is still open rather than leaving a stale verdict to merge past.

prog893:
@coderabbitai Requesting an explicit sign-off review. The MCP no-data finding was correct and is fixed in `496eaaa`.

You were right that this was the one place the two surfaces disagreed. `commands/get-channel-analytics.ts:78` emits the notice before its empty-rows check at line 82, while the MCP handler's empty-rows branch returned a bare string and dropped it. An affected pre-cutoff query with no rows informed a CLI user and told an MCP caller nothing.

Fixed by carrying the serialized `viewCountingNotice` alongside the no-data message. The reasoning, recorded in the code: the notice describes the requested **date range**, not the rows, so it still applies when nothing came back. A caller comparing an empty result against pre-cutoff archived data needs the same caveat as one comparing totals.

Verified against the live API rather than reasoned about, using a query that genuinely returns zero rows (`@staqan`, `day`/`views`, `2019-01-01..2019-01-05`):

```
before: No analytics data available for this channel and time period.

after:  No analytics data available for this channel and time period.

        { "viewCountingNotice": { "changeDate": "2026-08-24",
            "affectedMetrics": ["views"],
            "affectedRange": { "startDate": "2019-01-01", "endDate": "2019-01-05" },
            "spansChange": false, ... } }
```

Control: the same zero-row shape on a post-cutoff range (`2026-08-24`) carries no notice and returns the bare message, byte-identical to before. So the change adds the signal exactly where it was missing and nowhere else.

`youtube_get_video_analytics` has no empty-rows branch, so it already serialized the whole result and needed no change.

Thank you for withdrawing the parameter-object finding on the CLI-flag-vs-signature distinction.

`type-check`, `lint`, `build` and `bun test` (216 pass) are green. Nothing is outstanding from my side. Please post an updated review, and say if anything is still open rather than leaving a stale verdict to merge past.

Issue #185, opened 2026-08-24
Analytics: decide whether the fixed report shapes should carry engagedViews


prog893:
## Context

Split out of #175, which added `engagedViews` to `DEFAULT_VIDEO_METRICS` for `get-video-analytics`. Several other surfaces still send `views` as a fixed, non-overridable metric set.

The original reason for leaving them alone (recorded in #179) was that `engagedViews` compatibility for their dimensions had never been measured. **That reason no longer applies:** the #175 sweep measured it, and all three shapes below were probed directly and accept it.

## Measured 2026-08-24

| query shape | `views,engagedViews` |
|---|---|
| `insightTrafficSourceType` filtered to a video (`get-traffic-sources`) | accepted, 9 rows |
| `insightTrafficSourceDetail` + `YT_SEARCH` filtered to a video (`get-search-terms`) | accepted, 2 rows |
| `insightTrafficSourceDetail` + `YT_SEARCH` channel-wide (`get-channel-search-terms`) | accepted, 10 rows |

Plus every dimension used by `CHANNEL_REPORT_TYPES` (`country`, `deviceType`, `operatingSystem`, `insightTrafficSourceType`, `subscribedStatus`) permits it, per `docs/analytics-compat-snapshot.json`.

## The sites

- `lib/analytics.ts` `CHANNEL_REPORT_TYPES` — five predefined reports, four of which send `views,estimatedMinutesWatched`
- `lib/analytics.ts` `fetchTrafficSources` — sends `metrics: 'views'`
- `lib/analytics.ts` `fetchSearchTerms` — sends `metrics: 'views'`
- `lib/analytics.ts` `SEARCH_TERMS_METRICS` — `'views,estimatedMinutesWatched'`, used by the channel search-terms query

## Why this is a real decision, not a cleanup

It matters most on Shorts, where the two metrics diverge sharply (measured: 197 `views` against 8 `engagedViews`). A traffic-source or search-term breakdown that reports only `views` attributes reach rather than engagement, which is usually the question being asked.

Against that: each of these is a **fixed output shape** parsed by column, and there are five commands involved. Widening them all is a larger output-shape change than #175 made, and it affects `--report` types whose whole point is a stable shape.

Also note the sort interaction from #179: `CHANNEL_REPORT_TYPES` entries carry an explicit `sort: '-views'`. If a report gains `engagedViews`, which axis it should rank by becomes a live question rather than an inherited default.

## Proposed work

[...]

PR #186, opened 2026-08-24
feat(#175): add engagedViews, measured against the live API post-cutoff

### LINKED ISSUE #175: Analytics: add engagedViews support so pre-2026-08-24 view semantics stay queryable

prog893:
Closes #175.

## Why it needed measuring, not assuming

#175 item 1 says to verify against the live API after 2026-08-24, because `engagedViews` postdates the #173 sweep and that sweep's conclusions do not transfer. Ran on the cutoff date, the first day it could be measured post-change.

Added `engagedViews` to the sweep's candidate list and re-ran:

```
305 probes, 0 composition-law contradictions
engagedViews permitted by ALL 13 valid dimensions
```

The snapshot delta is clean: `validDimensions`, `invalidPairs` and `arityCaps` are **byte-identical** to the previous run, and the only change is `+engagedViews` on every dimension row. So it sits in the same compatibility class as `views`, including the five restrictive dimensions that permit nothing else, and joins `VIEW_METRICS`.

## It also demonstrably works

API acceptance is not the same as returning meaningful data, so measured over 2026-07-13..2026-08-23:

| video | type | `views` | `engagedViews` |
|---|---|---:|---:|
| BpesXOddpVc | short | 197 | **8** |
| qC_vD4tPNqA | short | 245 | **23** |
| 1zqV4DUboCY | short | 196 | **13** |
| -COxZI7L-IA | long-form | 548 | 548 |

Long-form is identical because a long-form view already counted from the start; the alignment removed a minimum-watch-time rule that only ever applied to Shorts. **That asymmetry is the argument for the default set.** A default query on a Short reported 197 where 8 is the engagement figure, so leaving it out was itself misleading.

## The default-set decision (#175 item 2)

#175 flagged widening the default as costly, citing #173: the default set already fails for most `--dimensions` values. **The sweep removes that objection specifically for this metric.** Unlike `likes`/`dislikes`/`comments`/`shares`, `engagedViews` is permitted everywhere, so it is never dropped and never narrows a query.

Verified end to end, `--dimensions deviceType`:

```
before:  deviceType,views,estimatedMinutesWatched,averageViewDuration,averageViewPercentage
         Note: dropped likes, dislikes, comments, shares ...

after:   deviceType,views,engagedViews,estimatedMinutesWatched,averageViewDuration,averageViewPercentage
         Note: dropped likes, dislikes, comments, shares ...   <- unchanged
```

Same four dropped, nothing else, plus the new column.

## E2E proof

Default query on a Short, main `7e7d674` before vs branch after:

```
before: video,views,estimatedMinutesWatched,...          -> 197, (no engagement figure)
after:  video,views,engagedViews,estimatedMinutesWatched,... -> 197, 8
```

9 columns to 10; every other column's value identical.

## Breaking-ish, and recorded

The added column changes default `json`/`csv`/`table` shape for callers parsing by position or column count. Recorded in `BREAKING-CHANGES.md` with the reasoning, the measurements, and the escape hatch (pass the previous eight explicitly). Callers reading by name, or passing `--metrics`, are unaffected.

## Docs (#175 items 3, 4, 5)

- `docs/dimension-compatibility.md`: matrix regenerated to nine columns and **verified cell-by-cell against the snapshot programmatically**, not by hand. Stale "all eight / permits four" counts corrected.
- `docs/commands/analytics.md`: metric added to both lists, plus a `views` vs `engagedViews` section carrying the Shorts table.
- `commands/mcp.ts`: metric schema description names both, states which answers which question, and gives the Shorts figures so an agent caller does not assume they agree.

## One correction carried forward

The `CHANNEL_REPORT_TYPES` comment (from #179) said `engagedViews` compatibility there was unswept. It is now swept, and those dimensions **do** accept it. Rather than leave a stale claim, the comment now states the remaining reason (fixed output shapes, five commands, a separate decision from this one) and the open question is tracked in **#185** with the measurements.

Tests 223 -> 229, including a guard that `views` and `engagedViews` stay in the same compatibility class so a future sweep breaking that parity fails loudly. `type-check`, `lint`, `build` green.

## Summary by CodeRabbit

* **New Features**
  * Added `engagedViews` as the second default metric for video analytics.
  * Made `engagedViews` available across supported analytics dimensions.
  * Clarified differences between `views` and `engagedViews`, including Shorts counting behavior.

* **Breaking Changes**
  * Default JSON, CSV, and table outputs now include an additional metric.
  * Consumers requiring the previous output shape must provide an explicit metric list.

* **Documentation**
  * Updated analytics guides, compatibility references, and tool descriptions with metric availability and definitions.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -16,8 +16,9 @@ by column position or column count. Callers that pass `--metrics` explicitly
 are unaffected, as are those reading columns by name.
 
 **Why:** on 2026-08-24 YouTube began counting a view from the first frame with
-no minimum watch time. `engagedViews` carries the stricter previous definition
-and is the figure tied to monetization. The two diverge sharply on Shorts,
+no minimum watch time. `engagedViews` carries the stricter previous definition,
+which is the only way to keep one definition across the cutoff. The two diverge
+sharply on Shorts,
 measured over 2026-07-13..2026-08-23:
 
 | type | `views` | `engagedViews` |
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -1,5 +1,47 @@
 # Breaking Changes
 
+## Unreleased
+
+**`get-video-analytics` default metrics gained `engagedViews`** (#175).
+
+The default set went from eight metrics to nine, with `engagedViews` inserted
+second, immediately after `views`:
+
+| Before | After |
+|---|---|
+| `views,estimatedMinutesWatched,...` | `views,engagedViews,estimatedMinutesWatched,...` |
+
+**Who is affected:** anything parsing default `--output json`, `csv` or `table`
+by column position or column count. Callers that pass `--metrics` explicitly
+are unaffected, as are those reading columns by name.
+
+**Why:** on 2026-08-24 YouTube began counting a view from the first frame with
+no minimum watch time. `engagedViews` carries the stricter previous definition
+and is the figure tied to monetization. The two diverge sharply on Shorts,
+measured over 2026-07-13..2026-08-23:
+
+| type | `views` | `engagedViews` |
+|---|---:|---:|
+| short | 197 | 8 |
+| short | 245 | 23 |
+| long-form | 548 | 548 |
+
+A default query on a Short reporting only `views` shows 197 where 8 is the
+engagement figure, so omitting it from the default was itself misleading.
+
+**To keep the old shape**, pass the previous set explicitly:
+
+```bash
+staqan-yt get-video-analytics --video-id ID \
+  --metrics views,estimatedMinutesWatched,averageViewDuration,averageViewPercentage,likes,dislikes,comments,shares
+```
+
+Adding it costs nothing in compatibility: the live sweep measured
+`engagedViews` as permitted by all 13 valid dimensions, so unlike
+`likes`/`comments`/`shares` it is never dropped when `--dimensions` is used.
+
+---
+
 ## v2.1.0
 
 **`--tags` → `--replace`** in `update-video-tags`:
--- a/commands/mcp.ts
+++ b/commands/mcp.ts
@@ -264,7 +264,7 @@ const TOOLS: Tool[] = [
         },
         metrics: {
           type: 'string',
-          description: 'Comma-separated list of metrics (default: views,engagedViews,estimatedMinutesWatched,averageViewDuration,averageViewPercentage,likes,dislikes,comments,shares). `views` counts every playback from the first frame; `engagedViews` applies the stricter pre-2026-08-24 definition and is the figure tied to monetization. On Shorts the two differ sharply (measured: 197 views against 8 engagedViews), so report the one the question calls for rather than assuming they agree. Both are accepted by every valid dimension.',
+          description: 'Comma-separated list of metrics (default: views,engagedViews,estimatedMinutesWatched,averageViewDuration,averageViewPercentage,likes,dislikes,comments,shares). `views` counts every playback from the first frame; `engagedViews` applies the stricter pre-2026-08-24 definition and is the only one consistent across that date. On Shorts the two differ sharply (measured: 197 views against 8 engagedViews), so report the one the question calls for rather than assuming they agree. Neither is a YouTube monetization figure: the Partner Programme uses its own separate metrics, so do not present either as earnings or eligibility. Both are accepted by every valid dimension.',
         },
         dimensions: {
           type: 'string',
--- a/commands/mcp.ts
+++ b/commands/mcp.ts
@@ -264,7 +264,7 @@ const TOOLS: Tool[] = [
         },
         metrics: {
           type: 'string',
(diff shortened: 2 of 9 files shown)


prog893:
@coderabbitai full review

prog893:
@coderabbitai Requesting an explicit sign-off review. Both findings were correct and are fixed in `23ace36`.

**1. `engagedViews` described as the monetization / YPP eligibility metric. Correct, and worth more than a wording fix.**

That claim was never verified. It came from the issue text and I repeated it into shipped documentation. Your distinction between earnings metrics and eligibility metrics is the kind of detail that makes a restated version just as unverifiable, so rather than qualify it I removed the claim from all six sites and replaced it with what was actually measured: `engagedViews` applies the stricter pre-2026-08-24 definition and is the only view metric comparable across that date.

Added explicit warnings rather than just deleting:

- `docs/commands/analytics.md` now states up front that the Partner Programme uses separate metrics and that `engagedViews` is not a payout or eligibility figure.
- `commands/mcp.ts` says the same to agent callers, since an agent asked about revenue is the most likely consumer to misuse this field.

Note two of the six sites predate this branch (`docs/commands/analytics.md` and the `SORTABLE_METRIC_PREFERENCE` comment, from #178 and #179, both already on `main`), plus one in `docs/troubleshooting.md`. Same false claim, so they are corrected here rather than left to spread.

**2. Stale metric guidance. Correct on both counts.**

`docs/commands/analytics.md` claimed `--dimensions deviceType` "fails until you pass `--metrics views`". As you note, it does not fail. It has not since #173 taught the CLI to drop unsupported defaults, so that passage was **already stale before this PR** and my new text contradicted it outright. Rewritten to describe the real behavior and to draw the distinction that actually matters: a defaulted metric list is narrowed automatically, an explicit one errors instead of being silently substituted.

`docs/dimension-compatibility.md` said the query returns "the four view and watch-time columns". It returns five.

Verified against the live API rather than reasoned about:

```
$ staqan-yt get-video-analytics --video-id BpesXOddpVc --dimensions deviceType ...
deviceType,views,engagedViews,estimatedMinutesWatched,averageViewDuration,averageViewPercentage
stderr: dropped likes, dislikes, comments, shares from the default metrics; not available for --dimensions deviceType.
```

Five columns, and the drop notice names exactly the four the prose now names.

`type-check`, `lint`, `build` and `bun test` (229 pass) are green. Nothing is outstanding from my side. Please post an updated review, and say if anything is still open rather than leaving a stale verdict to merge past.

prog893:
@coderabbitai Both findings from your review are fixed in `23ace36`: the unverified monetization/YPP claim is removed from all six sites (with an explicit warning added in its place), and the stale deviceType guidance now says five columns and describes automatic narrowing rather than failure, verified against the live API.

Requesting an explicit verdict. Nothing is outstanding from my side.

PR #187, opened 2026-08-24
docs(#176): caveat the Data API viewCount cutoff, and fix the jq recipes

### LINKED ISSUE #176: Data API viewCount changes definition on 2026-08-24; output and docs carry no caveat
## Context

On **2026-08-24** the `statistics.viewCount` field in the YouTube Data API switches to the new view-counting logic (counted from the first frame, no minimum watch time). The value under the previous logic will no longer be reachable through the public Data API at all: it is only available as Engaged Views through the Analytics and Reporting APIs.

The field name does not change, the type does not change, and nothing in the CLI output signals that the number after that date answers a different question than the number before it.

## Affected surfaces

- `lib/youtube.ts:276` and `lib/youtube.ts:403` parse `item.statistics.viewCount`
- `types/index.ts:69,268` declare `viewCount: number`
- `commands/get-video.ts:41,56,70,96` print and emit it as `views` / `viewCount`
- `commands/get-channel.ts:50,65,84,106` same for channel totals, labelled `Total Views`
- `docs/output-formats.md:110,119,145,158,321,338,342` and `docs/README.md:176-178` teach jq recipes that filter and sum `viewCount`

The concrete failure: any workflow that stored `viewCount` from `get-video --output json` before the cutoff and compares it to a value pulled after the cutoff measures the methodology change, not the video. The delta will be positive for every video, and nothing in either JSON payload distinguishes the two.

## Proposed work

1. Document the cutoff in `docs/commands/video-discovery.md`, `docs/commands/channel-operations.md` and `docs/output-formats.md`: state plainly that `viewCount` from the Data API is post-change from 2026-08-24 and that the previous definition is only available as `engagedViews` via Analytics (see the engagedViews issue).
2. Decide whether the JSON payload should carry a marker. A `viewCountMethodology` field, or a `_fetchedAt` timestamp that lets a consumer decide, would let downstream tooling detect the boundary without hardcoding a date. Adding a field is a breaking-ish change to output shape, so it belongs in `BREAKING-CHANGES.md` if taken.
3. Cross-check the jq recipes in the docs. Any that describe a historical comparison should point at Analytics rather than the Data API.

## References

- [videos.list statistics.viewCount](https://developers.google.com/youtube/v3/docs/videos#statistics.viewCount)
- [Data API revision history](https://developers.google.com/youtube/v3/revision_history)

prog893:
Closes #176. Docs only, no source changes.

## The caveat (#176 item 1)

`statistics.viewCount` switched to the new view-counting logic on 2026-08-24. The field name and type did not change, so a payload spanning the cutoff gives a consumer no signal that it now answers a different question, and the previous definition is **not reachable through the Data API at all** (only as `engagedViews` via Analytics).

Added where the field is surfaced:

- **`video-discovery.md`** (`get-video`, `get-videos`)
- **`channel-operations.md`**, where the same number is labelled `Total Views`. Worth calling out separately: that one is a *lifetime* total, so it necessarily sums across both definitions rather than merely risking it.
- **`output-formats.md`**, scoped to the recipes that aggregate or compare rather than merely display.

## No payload marker (#176 item 2) — decided against

The issue floated `viewCountMethodology` or `_fetchedAt`. Declining both:

- `viewCountMethodology` would be a **constant** post-cutoff, carrying nothing a doc line does not, and it cannot retroactively mark data already archived.
- `_fetchedAt` would make **every payload differ from every other**, breaking the diff and checksum workflows `output-formats.md` itself recommends, and which this repo's own E2E merge gate depends on.

The reasoning is written into the docs rather than left implicit, so it does not get re-proposed later. No `BREAKING-CHANGES.md` entry is needed, since output shape is unchanged.

## The jq cross-check (#176 item 3) found something bigger

The issue asks to cross-check the recipes for historical-comparison advice. They turned out to be **wrong on their own terms**, pre-existing and unrelated to the date:

| documented | actual |
|---|---|
| `list-videos ... jq '.[].viewCount'` (4 recipes) | `list-videos` returns **no statistics at all** |
| top-level `.viewCount` | nested at `.statistics.viewCount` |
| `.viewCount \| tonumber`, sample `"1400000000"` | it is already a **number** |
| `csvsort --columns viewCount` | the CSV column is named **`views`** |
| csv sample `id,title,viewCount,publishedAt` | `id,title,channel,published,duration,views,likes,comments,type` |
| text sample `id\ttitle\t1400000000` | `id\ttitle\tdate\tprivacy\ttype` |

So `csvsort --columns viewCount` could never have worked, and the four `list-videos` recipes returned `null` for every row.

**Every rewritten recipe was run against the live CLI before committing**, not reasoned about:

```
get-videos --video-ids ... | jq 'select(.statistics.viewCount > N)'   -> {"id":"BpesXOddpVc","v":197}
jq -r '.[0].title'  (result is an array)                             -> ok
jq '[.[].statistics.viewCount] | add'                                -> 197
jq -r '.[] | [.id,.title,.statistics.viewCount] | @csv'              -> ok
csvsort --columns views                                              -> column exists
get-video-analytics --metrics views,engagedViews                     -> 197, 8
```

Samples now show the headers the commands actually print, and the caveat names where the field lives per format (`statistics.viewCount` in json, `views` column in csv/table) so the mapping is not guesswork.

## One recipe I wrote, tested, and had to replace

I first documented `get-channel-analytics --dimensions video --metrics views,engagedViews --sort engagedViews`. It fails. Probing the API directly:

| query | result |
|---|---|
| `dimensions=video` + metrics, no sort | rejected |
| `+ sort=-engagedViews` | rejected |
| `+ sort=-engagedViews` **+ maxResults** | **OK** |
| `+ sort=engagedViews` (ascending) + maxResults | rejected |

So that report requires `maxResults` and permits descending sort only. `fetchChannelAnalytics` never sends `maxResults`, which means **`get-channel-analytics --dimensions video` cannot work today** regardless of metrics. Replaced with a verified `get-video-analytics` recipe; the missing-`--limit` gap is left unfiled pending your call on whether it is worth a separate issue.

## Relationship to #186

Independent. This branch is off `main` and touches no source, so it does not depend on the `engagedViews` default-metrics change in #186 and will not conflict with it. The recipes here pass `--metrics` explicitly, so they work on `main` today either way.

`type-check`, `lint`, `build`, 223 tests green.

## Summary by CodeRabbit

* **Documentation**
  * Updated guidance for YouTube’s August 24, 2026 `viewCount` methodology change and mixed historical totals.
  * Recommended Analytics `engagedViews` for evaluating engagement under the former definition.
  * Clarified which commands provide statistics and how to retrieve video statistics before filtering, sorting, or analyzing performance.
  * Revised command examples and JSON, CSV, and text output documentation to reflect nested statistics fields and updated column names.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -173,9 +173,24 @@ cat video_ids.txt | \
 ### Find Underperforming Videos
 
 ```bash
-# Get videos with low views
-staqan-yt list-videos @mychannel --limit 100 --output json | \
-  jq '.[] | select(.viewCount | tonumber < 1000) | {title, viewCount}'
+# Get videos with low views.
+# Counts come from get-videos, not list-videos: list-videos returns no
+# statistics at all. Pass the IDs you want, or collect them from list-videos.
+staqan-yt get-videos --video-ids ID1 ID2 ID3 --output json | \
+  jq '.[] | select(.statistics.viewCount < 1000) | {title, views: .statistics.viewCount}'
+```
+
+Since 2026-08-24 `viewCount` counts every playback from the first frame, which
+inflates Shorts relative to a threshold chosen before that date. To judge a
+video by engagement instead, pull `engagedViews`, which keeps the stricter
+definition:
+
+```bash
+staqan-yt get-video-analytics --video-id VIDEO_ID \
+  --metrics views,engagedViews --output csv
+
+# video,views,engagedViews
+# BpesXOddpVc,197,8      <- a Short: 197 reached, 8 engaged
 ```
 
 ### Archive Thumbnail CTR Data
--- a/docs/commands/channel-operations.md
+++ b/docs/commands/channel-operations.md
@@ -227,8 +227,11 @@ staqan-yt list-videos @yourchannel --limit 20 --output table
 
 ```bash
 # Export video stats for analysis
-staqan-yt list-videos @yourchannel --output json | \
-  jq '.[] | {title, viewCount: .viewCount | tonumber, likeCount: .likeCount | tonumber}' | \
+# Counts come from get-videos: list-videos returns no statistics.
+# Collect the IDs first, then fetch them in one batch.
+staqan-yt list-videos @yourchannel --output json | jq -r '.[].id' | \
+  xargs staqan-yt get-videos --output json --video-ids | \
+  jq '.[] | {title, viewCount: .statistics.viewCount, likeCount: .statistics.likeCount}' | \
   jq -s 'sort_by(.viewCount) | reverse'
 ```
 
--- a/docs/commands/channel-operations.md
+++ b/docs/commands/channel-operations.md
@@ -2,6 +2,14 @@
 
 Commands for working with YouTube channels and their content.
 
+> **`viewCount` (shown as `Total Views`) changed meaning on 2026-08-24.** From
+> that date the Data API counts a view from the first frame of playback, with
+> no minimum watch time. A channel total spanning the cutoff is therefore a sum
+> over two different definitions, and comparing today's total against one
+> stored earlier measures the methodology change as well as the channel. The
+> previous definition survives only as `engagedViews` in the Analytics API.
+> See [The 2026-08-24 view-counting change](analytics.md#the-2026-08-24-view-counting-change).
+
 ## get-channel
 
 Get detailed metadata for a YouTube channel.
--- a/docs/commands/channel-operations.md
+++ b/docs/commands/channel-operations.md
@@ -2,7 +2,7 @@
 
 Commands for working with YouTube channels and their content.
 
-> **`viewCount` (shown as `Total Views`) changed meaning on 2026-08-24.** From
+> **`statistics.viewCount` (shown as `Total Views`) changed meaning on 2026-08-24.** From
 > that date the Data API counts a view from the first frame of playback, with
 > no minimum watch time. A channel total spanning the cutoff is therefore a sum
 > over two different definitions, and comparing today's total against one
@@ -52,12 +52,16 @@ staqan-yt get-channel @mkbhd --output json
 - `id` - Channel ID
 - `title` - Channel name
 - `handle` - Channel handle (@username)
+- `customUrl` - Custom channel URL
 - `description` - Channel description
+- `country` - Channel country code
 - `publishedAt` - Channel creation date
-- `thumbnail` - Channel thumbnail URL
-- `subscriberCount` - Subscriber count
-- `videoCount` - Total video count
-- `viewCount` - Total lifetime view count
+- `brandingSettings` - Channel branding (banner, keywords, etc.)
+- `topicDetails` - Topic categories
+- `statistics` - Object containing `subscriberCount`, `videoCount`, `viewCount`, `hiddenSubscriberCount`
+
+The counts are **nested under `statistics`**, so the jq path for the lifetime
+total is `.statistics.viewCount`, not `.viewCount`.
 
 ### Related Commands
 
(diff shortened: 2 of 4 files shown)


prog893:
@coderabbitai full review

Context worth flagging for the review: this is a docs-only change. Beyond the 2026-08-24 viewCount caveat the issue asked for, cross-checking the jq recipes turned up that several were already broken independently of the date (reading .viewCount off list-videos, which returns no statistics at all; a top-level path where the field is nested under .statistics; and a csv column named views, not viewCount). Every rewritten recipe was run against the live CLI before committing.

prog893:
@coderabbitai Requesting an explicit sign-off review. Both findings were correct and are fixed in `10c3649`, and the second one led to two more instances of the same bug.

**1. `docs/output-formats.md:348`, reading statistics from `list-videos`. Correct, and an embarrassing miss.**

I corrected that recipe's jq projection to `.statistics.viewCount` but left its source command as `list-videos`, which is the exact bug this PR exists to fix. The page then contradicted itself, since it states a few lines above that `list-videos` carries no statistics. Now sourced from `get-videos`.

**Because I missed one while fixing that class of bug, a second manual pass was not trustworthy.** I checked the rest by script instead: every `list-videos ... --output json` whose following lines read `.statistics` / `.viewCount` / `.likeCount` without handing off to `get-videos` first. That found **two more**, both in files this branch had already touched and neither flagged in review:

| file | recipe | was |
|---|---|---|
| `docs/commands/channel-operations.md` | "Analyze Video Performance" | `.viewCount`, `.likeCount` off `list-videos` |
| `docs/commands/video-discovery.md` | "Find Most Viewed Videos" | `sort_by(.viewCount)` off `list-videos` |

Both had been emitting `null` for every row. Rewritten to collect IDs from `list-videos` and batch them through `get-videos`, which is the only shape that yields counts.

All three run against the live CLI, not reasoned about:

```
channel-operations  -> 557/14, 245/2, 197/2   (sorted descending)
video-discovery     -> 557, 245, 197, 196     (sorted descending)
combining-formats   -> "BpesXOddpVc","You make home wh",197
```

The script reports clean across `docs/` now.

**2. `docs/output-formats.md:6`, ambiguous heading. Correct.**

"not comparable across 2026-08-24" can be read as comparing within that single date. Restated as before-versus-on-or-after, which is the actual boundary.

`type-check`, `lint`, `build` and `bun test` (223 pass) are green. Docs-only, no source changes. Nothing is outstanding from my side. Please post an updated review, and say if anything is still open rather than leaving a stale verdict to merge past.

prog893:
@coderabbitai Requesting an explicit sign-off review. Your finding was correct and is fixed in `8b5b75e`, and checking its siblings found the same defect in three more places.

**`list-videos` "Output Fields" documenting statistics it does not return.** Correct, and it contradicted the corrected recipe a few sections below in the same file.

This was the third instance of this class of error in this branch, so I checked the sibling blocks by script rather than by eye. **All four were wrong** — 11 documented fields that do not exist, 8 real fields undocumented:

| command | ghost (documented, not returned) | missing (returned, undocumented) |
|---|---|---|
| `list-videos` | `channelId` `channelTitle` `viewCount` `likeCount` `commentCount` `duration` | `videoType` `privacyStatus` |
| `get-video` | `channelId` `thumbnail` `viewCount` `likeCount` `commentCount` | `thumbnails` `videoType` `privacyStatus` `categoryId` `statistics` |
| `get-channel` | `thumbnail`; plus `subscriberCount`/`videoCount`/`viewCount` listed top-level when nested | `customUrl` `country` `brandingSettings` `topicDetails` `statistics` |
| `search-videos` | none | `videoType` |

The single pattern behind all of them: counts live under a nested `statistics` object on `get-video`, `get-videos` and `get-channel`, and **do not exist at all** on `list-videos` or `search-videos`. Every block now says which, and the two carrying statistics state the jq path explicitly, since a top-level `.viewCount` yields `null` silently rather than failing.

Worth noting the caveat this PR adds tripped over the same defect: `channel-operations.md` introduced the cutoff as affecting `viewCount`, when the field is only reachable as `statistics.viewCount`. Corrected, so the caveat now names the path a reader can actually use.

Field lists verified against live output for all four commands, and the checking script reports every block accurate.

You also confirmed no further `list-videos --output json` recipe reads statistics without an intervening `get-videos`, which matches my own sweep.

`type-check`, `lint`, `build` and `bun test` (223 pass) are green. Docs-only. Nothing is outstanding from my side. Please post an updated review, and say if anything is still open rather than leaving a stale verdict to merge past.

Issue #106, opened 2026-07-02
Cache: handle index version migration (delete orphaned files on mismatch)


prog893:
## Problem

When cache index version doesn't match current code, the code warns and returns an empty index but doesn't actually migrate or delete old data. Old cached files remain orphaned on disk indefinitely, taking up space and confusing users about what's actually cached.

## Current State

`lib/cache.ts` checks version, warns on mismatch, returns empty index. Old files (potentially GB of cached reports) stay in `~/.staqan-yt-cli/data/{channelId}/`.

## Proposed Solution

1. **On version mismatch, offer migration**:
   - Auto-migrate if format is forward-compatible
   - Otherwise, move old cache to `~/.staqan-yt-cli/data/.archive/{timestamp}/`
2. **Add `cache clean` command** for manual cleanup of orphaned files
3. **Document cache format version** in `lib/cache.ts` header for future migrations
4. **Track cache size** in `cache status` output

## References

- Documented in `staqan/research/staqan-yt-cli-architecture.md` (file 61) as "Potential Issue #6: Cache index migration"
- Related closed: #38 "Improve cache save failure handling and add cache cleaning command" (partially addresses this)

prog893:
**Scope addition (2026-07-10 audit):** the version-mismatch warning in `lib/cache.ts` currently says *"Cached report data cleared."* — but only the **index** is reset; the report CSVs and metadata files stay on disk as orphans. The message asserts the opposite of what happens, which will actively mislead whoever debugs a migration. Whatever shape the migration fix takes, correcting that message (or making it true by actually archiving the files, per the `.archive/{timestamp}/` proposal) should be part of the acceptance criteria.

PR #189, opened 2026-08-24
fix(#177): warn when merged bulk reports span the view-counting change

### LINKED ISSUE #177: Reporting cache: merged bulk reports mix pre/post 2026-08-24 views and a new engaged_views column

prog893:
Closes #177.

## The premise in the issue does not hold

The issue proposes detecting this by comparing report schemas, on the premise that Bulk Reports **gain** an `engaged_views` column at the 2026-08-24 cutoff. Measured against the local archive (3,683 reports, 20 report types, windows spanning 2026-01-22 to 2026-08-22):

| | result |
|---|---|
| report types carrying `views` that also carry `engaged_views` | **all of them** |
| earliest archived window already carrying `engaged_views` | **2026-01-22** |
| report types with more than one column set | **0** |

There is no new column. `engaged_views` has been in every affected report type since long before the cutoff, and no report type has ever shown schema drift. A schema comparison would have fired on nothing while the actual defect went unreported.

Also worth recording: `ReportMetadata.columns` is written to the **per-report metadata sidecar**, not to the cache index. All 3,683 index entries lack a `columns` key, so the issue's "detectable without a new fetch" is true only via the sidecars, which is the path this PR uses.

## The actual defect

The half of the issue that does hold is the serious half: **`views` changes meaning on 2026-08-24 while keeping its name.** The header row is byte-identical on both sides of the date. A range straddling the cutoff merges rows measured two different ways into one column, and summing it yields a number that counts nothing in particular. This is invisible to any schema check and visible only in the dates.

That makes the bulk path more exposed than the Analytics path, not less: an archive accumulates for months, so a single `get-report-data` call routinely spans the cutoff by default.

## What this changes

**A date-keyed notice on the reporting path**, extending the existing #178 machinery rather than adding a parallel one. The builder is now parameterized by field vocabulary, so each surface reports the spelling it actually emits:

| surface | affected fields | consistent field |
|---|---|---|
| Analytics API | `views`, `redViews` | `engagedViews` |
| Bulk Reporting API | `views`, `red_views` | `engaged_views` |

The correspondence is stated rather than derived, because it is not mechanical: `estimatedMinutesWatched` is `watch_time_minutes`.

The notice prints for **every** output format here, unlike `get-video-analytics` and `get-channel-analytics`, which gate theirs on pretty/text. That follows this command instead of those two: its Incomplete Data warning is the same class of caveat and is already unconditional, so gating this one would leave the weaker caveat more visible than the stronger. stderr keeps stdout parseable either way.

**A `schemaMismatch` signal** answers the part of the issue that stands. If a report type is ever revised, the formatters union the keys and emit a column blank for some dates, indistinguishable from a genuine empty value. Reporting it is the only way a consumer can tell those apart. It returns undefined on all of today's data; it guards a future revision.

On the issue's item 2 (what a missing key should carry): `formatCsv` already unions keys and emits an empty cell for a missing one, so the CSV shape is stable today and changing the row shape would be a silent data change with no live instances to justify it. The mismatch signal is the answer to distinguishability. Item 4 (a schema-version field) stays with #106.

Both fields reach MCP callers through the serialized result; the tool description now says what to do with them.

## E2E proof

Identical command, real archive, `channel_basic_a3`, `2026-05-01` to `2026-08-24`:

| | main (`4473778`) | this branch |
|---|---|---|
| exit code | 0 | 0 |
| stdout bytes | 156081 | 156081 |
| stdout sha256 (first 16) | `6376cb9726edaa63` | `6376cb9726edaa63` |
| view-counting lines on stderr | **0** | **1** |

stdout is byte-identical, so no data regression: the warning is stderr-only.

```
Note: views, red_views for 2026-05-01 to 2026-08-22 is counted under the definition
used before 2026-08-24, so it is not comparable to views, red_views for dates from
2026-08-24 onward, which counts every playback from the first frame. engaged_views
keeps the stricter definition throughout.
```

Controls, all against the live archive:

- **No view columns stays silent.** `channel_demographics_a1` (columns end `age_group,gender,views_percentage`) emits no notice. This also proves the substring guard on real data: `views_percentage` contains "views" and does not trigger.
- **Machine output stays clean.** `--output json` parses, 237 rows, no notice text anywhere in stdout; the notice is on stderr.
- **Invalid input still rejected.** `--start-date not-a-date` exits 1 with the format error.
- **Post-cutoff boundary.** `--start-date 2026-08-24` still exits 1 with the pre-existing no-overlap error (that day has no data yet) and emits no spurious notice.
- **Analytics path unregressed.** `get-video-analytics` over the same range still says `engagedViews` in camelCase and still names only `views`, matching its pre-refactor output exactly.

Full suite 229 -> 246, type-check and lint clean.

## Summary by CodeRabbit

- **New Features**
  - Added view-counting notices to report results when data spans YouTube’s August 24, 2026 measurement change.
  - Notices identify affected columns, including `views` and `red_views`, and recommend `engaged_views` for comparisons across the change date.
  - Notices appear consistently across supported output formats.

- **Documentation**
  - Updated reporting and MCP documentation with guidance on interpreting mixed historical definitions, report versions, CSV output, and notification behavior.


prog893:
## Dropped the schema-mismatch guard (`9e0f304`)

The guard cannot fire, and shipping it was inconsistent with this PR's own finding.

`fetchReportData` is scoped to a single `--type` end to end: the job lookup (`lib/reports.ts:646`) and all three cache lookups (727, 788, 863) filter on `params.type`, and each type gets its own cache directory. Every report merged into one result is the same report type by construction, so a future `channel_basic_a4` would be a different ID, a different job, a different directory and a different query. It can never land in the same merge as `_a3`.

That leaves only one path to disagreeing columns: YouTube mutating a schema under a stable report ID. The revision history shows it does not. `engaged_views` arrived on **2025-06-24** as a version bump, `channel_basic_a2` to `channel_basic_a3`:

> "Updated reports will include a new column, `engaged_views`, which will reflect the previous view-counting methodology."

This also strengthens the original finding rather than just failing to contradict it. The archive measurement showed no schema drift since 2026-01, but could not rule out an in-place edit before that window. The revision history rules it out directly: the column shipped with a new ID, which is the convention.

A corrupted local CSV is already covered separately, by `readCachedReport` comparing parsed headers against the sidecar metadata and dropping the report on mismatch.

**Removed:** `findSchemaMismatch`, `ReportSchemaMismatch`, the `schemaMismatch` result field, the `progress()` warning, the CLI warning block, the sentence in the MCP tool description, and `tests/reports.test.ts` (it covered nothing else). `contributingColumns` stays, simplified to `string[][]`, since the notice still needs to know whether the result carries a view column.

**Also added:** the view-counting section in `docs/commands/reporting-api.md`. That file had zero mentions of the change or of `engaged_views` across 550 lines, while `docs/commands/analytics.md` carried a full section. The bulk surface is where reports merge across the cutoff, so it was the more important of the two to document.

Tests 246 -> 237 (the 9 removed cover the deleted function). What remains is the view-counting notice, which addresses the defect that is real: same column name, changed meaning, detectable only by date.

prog893:
@coderabbitai The single finding on this PR does not hold. Declining it, with evidence.

**1. The cited guideline governs CLI flags, not TypeScript signatures.**

It lives under `### 3. AWS API Naming Conventions` in CLAUDE.md, and every example under it is a shell invocation:

```
**🚨 RULE: Everything is a named flag. No positional arguments at all.**

**Singular = Single-item operations:**
​```bash
update-video --video-id <id>     # Update ONE video
​```
```

The same section in `docs/development/architecture.md#aws-api-naming-conventions` is likewise entirely about command naming and `--flag` shapes. It is the rule that forbids `staqan-yt get-video <id>` in favour of `staqan-yt get-video --video-id <id>`. It says nothing about internal function parameters.

**2. The codebase convention is the opposite of the proposed change.**

Counting exported functions in `lib/`:

| style | count |
|---|---|
| 2+ positional parameters | **29** |
| params/flags object | 8 |

The 8 using an object are all top-level fetchers with many optional fields (`fetchVideoAnalytics`, `fetchChannelAnalytics`, `fetchReportData`, and so on). Every pure helper is positional: `readCachedReport(channelId, reportId, reportTypeId)`, `findCachedReports(channelId, type, start, end)`, `loadReportMetadata(channelId, reportId, reportTypeId)`, `compareInstants(a, b)`, `resolveCustomSort(dimensions, metrics, explicit)`.

**3. The change would make the two notice builders inconsistent with each other.**

`viewCountingNoticeForColumns` is a deliberate mirror of `viewCountingNoticeFor(startDate, endDate, metrics)`, which shipped in #178 and is unchanged here. They are the same function for two field vocabularies and share `buildViewCountingNotice`. Converting only the new one leaves two sibling helpers with divergent call shapes, which is a worse readability outcome than the one the finding is aimed at.

All three parameters are required, and the pair the finding is most concerned about (`startDate`, `endDate`) appears in that order in dozens of signatures across this codebase.

Please confirm whether this finding is withdrawn, and state an explicit verdict for the PR.

prog893:
## Trial merge with #190: conflict in `tests/analytics.test.ts`

Both PRs branch from `4473778` and both append a `describe` block immediately after the `tolerates whitespace in the metric list` case, so they collide. CI cannot catch this, since neither PR has ever been built against the other.

```
Auto-merging lib/analytics.ts
Auto-merging tests/analytics.test.ts
CONFLICT (content): Merge conflict in tests/analytics.test.ts
```

`lib/analytics.ts` auto-merges cleanly. The test file needs a manual resolution, and git splits both blocks mid-body rather than at their boundaries, so the seam needs repairing (close the first `describe`, reopen the second block's doc comment) rather than a plain take-both.

Resolved locally with both blocks kept:

| | result |
|---|---|
| type-check | clean |
| lint | clean (4 pre-existing warnings) |
| tests | **241 pass, 0 fail** (237 from this PR + 4 from #190) |

Both features verified against the live API in the combined build:

- `get-report-data` emits `engaged_views keeps the stricter definition throughout` (snake_case, bulk surface)
- `get-traffic-sources` emits `engagedViews keeps the stricter definition throughout` (camelCase, Analytics surface)

The two vocabularies stay correctly separated when both are present. Whichever PR merges second will need this resolution.

Filed while investigating the cache metadata layer: #191 (sidecar read amplification, 482 reads for 114 reports), #192 (corrupt sidecar silently disables CSV validation), #193 (`startTimeActual` stored in two formats).

PR #190, opened 2026-08-24
fix: emit the view-counting notice for the fixed-metric reports

prog893:
Three commands return `views` over a date range and say nothing about the 2026-08-24 view-counting change, while `get-video-analytics` and `get-channel-analytics` have warned since #178.

| command | metrics sent | notice before | after |
|---|---|---|---|
| `get-traffic-sources` | `views` | no | yes |
| `get-search-terms` | `views` | no | yes |
| `get-channel-search-terms` | `views,estimatedMinutesWatched` | no | yes |

## Why this one is not theoretical

`get-traffic-sources` exposes **no date flags at all**. It hardcodes a rolling 30-day window, so for the first 30 days after the change its one and only behaviour straddles it. Run on 2026-08-25 against `main`:

```
Date Range: 2026-07-26 to 2026-08-25          (change date: 2026-08-24)
```

Two counting definitions in one `views` column, sorted by `-views`, with no caveat. No unusual invocation required to reach it.

These three also have less room to work around the ambiguity than the custom-query commands do: each sends a fixed metric set the caller cannot widen, so there is no `--metrics engagedViews` escape.

## Relationship to #185

#185 asks whether these reports should also **return** `engagedViews`. That is an output-shape decision across five commands, it interacts with the `sort: '-views'` axis, and it is unresolved.

This PR does not touch it. The notice goes to stderr and changes no output shape, so it is not blocked on that decision and does not pre-empt it. It just stops the commands being silent while the decision waits.

## Convention followed

Matches the two commands that already warn: stderr, `--output pretty` and `--output text` only, so `json`/`csv`/`table` stay clean for piping. MCP callers read `viewCountingNotice` off the result.

This is deliberately the opposite choice from the one in #189, where the notice is unconditional. That command's sibling Incomplete Data warning is already unconditional, so gating there would have left the weaker caveat more visible; these three have no such sibling and follow their own surface's precedent.

## E2E proof

`get-traffic-sources`, identical invocation, live API, main (`4473778`) vs branch:

| | main | branch |
|---|---|---|
| exit code | 0 | 0 |
| `--output text` stdout sha256 | `eea9de23f6141657` | `eea9de23f6141657` |
| `--output json` stdout sha256 | `5bb54c049c3851f9` | `5bb54c049c3851f9` |
| notice on stderr, `text` | **0** | **1** |
| notice on stderr, `json` | 0 | **0** (correctly gated) |

stdout byte-identical in both formats, so no data regression.

```
Note: views changes definition inside this date range. 2026-07-26 to 2026-08-23 is
counted under the previous definition; 2026-08-24 onward counts every playback from
the first frame. Totals across the whole range mix the two. engagedViews keeps the
stricter definition throughout.
```

`get-channel-search-terms` over `2026-05-01` to `2026-08-21` likewise went from 0 notices to 1, and correctly names only `views`, not `estimatedMinutesWatched`.

Tests 229 -> 233, including a case pinning that the rolling window goes quiet again once it clears the change date. Type-check and lint clean.

## Summary by CodeRabbit

* **New Features**
  * Added view-counting notices to traffic-source, search-term, channel search-term, and per-video reports when metric definitions may affect results.
  * Notices account for reporting date ranges, mixed metric definitions, and rolling-window changes.

* **Bug Fixes**
  * Human-readable `pretty` and `text` output now displays notices on stderr without disrupting report data.
  * JSON and CSV output remain unchanged, while MCP results continue exposing the notice as metadata.

* **Tests**
  * Added coverage for affected metrics, boundary dates, mixed definitions, and automatic notice removal.

### DIFF
--- a/commands/get-channel-analytics.ts
+++ b/commands/get-channel-analytics.ts
@@ -2,7 +2,7 @@ import chalk from 'chalk';
 import { debug, formatNumber, initCommand, withSpinner, validateDateOption, validateDateRange, runOrExit, writeStdout } from '../lib/utils';
 import { requireChannel, getOutputFormat } from '../lib/config';
 import { formatJson, formatTable, formatCsv } from '../lib/formatters';
-import { fetchChannelAnalytics, ChannelAnalyticsResult } from '../lib/analytics';
+import { fetchChannelAnalytics, ChannelAnalyticsResult, emitViewCountingNotice } from '../lib/analytics';
 import { ChannelAnalyticsOptions } from '../types';
 
 async function getChannelAnalyticsCommand(options: ChannelAnalyticsOptions): Promise<void> {
@@ -75,9 +75,7 @@ async function getChannelAnalyticsCommand(options: ChannelAnalyticsOptions): Pro
     // and a note they cannot parse is noise there; the same signal reaches
     // those callers as `viewCountingNotice` in the MCP result. stderr, not
     // stdout, so even here it never lands in a redirected file.
-    if (result.viewCountingNotice && (outputFormat === 'pretty' || outputFormat === 'text')) {
-      process.stderr.write(`${result.viewCountingNotice.message}\n`);
-    }
+    emitViewCountingNotice(result.viewCountingNotice, outputFormat);
 
     if (rows.length === 0) {
       console.log('');
--- a/commands/get-channel-search-terms.ts
+++ b/commands/get-channel-search-terms.ts
@@ -7,6 +7,7 @@ import {
   validateContentType,
   ALL_TIME_START_DATE,
   CHANNEL_SEARCH_TERMS_MAX_VIDEOS,
+  emitViewCountingNotice,
 } from '../lib/analytics';
 import { ChannelSearchTermsOptions } from '../types';
 
@@ -55,9 +56,7 @@ async function getChannelSearchTermsCommand(options: ChannelSearchTermsOptions):
     // get-channel-analytics: json/csv/table are what scripts read and a note
     // they cannot parse is noise there. MCP callers read `viewCountingNotice`
     // off the result instead.
-    if (result.viewCountingNotice && (outputFormat === 'pretty' || outputFormat === 'text')) {
-      process.stderr.write(`${result.viewCountingNotice.message}\n`);
-    }
+    emitViewCountingNotice(result.viewCountingNotice, outputFormat);
     const { channelId, channelTitle, videosAnalyzed, columnHeaders, rows } = result;
     const { startDate, endDate } = result.dateRange;
     const isLifetime = startDate === ALL_TIME_START_DATE;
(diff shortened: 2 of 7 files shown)


prog893:
@coderabbitai full review

The first review attempt hit the OSS quota before it ran, so the commits are marked seen and an incremental review would no-op. Requesting a full pass.

Context for the review: this PR adds the #178 view-counting notice to the three commands that send a fixed `views` metric set (`get-traffic-sources`, `get-search-terms`, `get-channel-search-terms`) and previously said nothing about the 2026-08-24 counting change. It also carries a merge of main resolving a conflict with #189 in `tests/analytics.test.ts`, where both PRs appended a describe block at the same anchor.

prog893:
@coderabbitai Finding accepted, addressed in `8e855b5`, though not by the route suggested.

The gap was real: `tests/analytics.test.ts` covered the notice builder, and nothing covered the wiring that decides which stream carries it. Deleting the emit block from any one of the three commands would have left the whole suite green.

**Why not mocked fetchers.** Mocking each fetcher also means stubbing `getAuthenticatedClient` and the googleapis client beneath it. This repo has no mocking infrastructure at all (no `mock.module`, no `spyOn` anywhere in `tests/`), so that route introduces a harness in a PR whose subject is a warning message. The existing precedent for stream-level assertions is `tests/stdout-flush.test.ts`, which spawns real child processes, and that route needs live credentials CI does not have.

**What was done instead.** The three new commands were repeating a rule that two existing commands had already written out by hand, five copies in total:

```ts
if (result.viewCountingNotice && (outputFormat === 'pretty' || outputFormat === 'text')) {
  process.stderr.write(`${result.viewCountingNotice.message}\n`);
}
```

That is extracted to `emitViewCountingNotice(notice, outputFormat, stream = process.stderr)`, which takes an injectable stream. The routing decision is now testable directly, and there is one copy of it instead of five. Five copies of "which stream carries this" is five chances for one to drift onto stdout, where it corrupts a piped result rather than merely annoying the reader.

Five tests added:

| case | asserted |
|---|---|
| `pretty`, `text` | returns true, writes the message once |
| `json`, `csv`, `table` | returns false, writes nothing |
| no notice | returns false, writes nothing |
| unknown format (`yaml`) | returns false, fails closed |
| default argument | lands on stderr, **stdout untouched** |

The last one pins the default parameter, since that is where the actual routing decision lives.

`get-report-data` deliberately keeps its own inline call rather than adopting the helper. It prints for every output format, alongside its unconditional Incomplete Data warning, which is a different rule. Making that a boolean flag on the helper would have hidden two policies behind one call.

**Behaviour verified unchanged against the live API**, since this refactors five call sites including two outside this PR's original scope:

| check | result |
|---|---|
| `get-traffic-sources --output text` stdout sha256 | `eea9de23f6141657`, unchanged |
| `get-traffic-sources --output json` stdout sha256 | `5bb54c049c3851f9`, unchanged |
| `get-video-analytics --output text` | still emits |
| `get-report-data --output csv` | still emits, confirming it stayed ungated |

Tests 233 -> 246. Type-check and lint clean.

Please re-review and state an explicit verdict.

PR #199, opened 2026-08-24
feat(#185): add engagedViews to the fixed-shape analytics reports

### LINKED ISSUE #185: Analytics: decide whether the fixed report shapes should carry engagedViews

prog893:
Closes #185.

## The decision

**Widen all four surfaces.** `engagedViews` goes immediately after `views`; `sort` stays `-views`.

#185 asked for a per-surface decision rather than a global one. Having measured each, they all land the same way, for the same reason.

## Measured live, 2026-08-25

#179 deferred this because compatibility was unmeasured. Probed every shape directly:

| shape | `views,engagedViews` |
|---|---|
| `insightTrafficSourceType` filtered to a video | accepted, 9 rows |
| `insightTrafficSourceDetail`+`YT_SEARCH` filtered to a video | accepted |
| `insightTrafficSourceDetail`+`YT_SEARCH` channel-wide | accepted, 10 rows |
| `deviceType,operatingSystem` / `country` / `insightTrafficSourceType` / `subscribedStatus` | all accepted |
| `sort: -engagedViews` | also accepted |

The same run shows why it matters. Channel-wide over `2026-05-01..2026-08-25`:

| report | `views` | `engagedViews` | gap |
|---|---:|---:|---:|
| geography | 1485 | 883 | 41% |
| devices | 2849 | 1873 | 34% |
| traffic-sources | 2873 | 1896 | 34% |
| channel search terms | 189 | 115 | 39% |

Per-video long-form the two agree exactly (548 against 548). The channel-level gap is the Shorts mix. So a breakdown reporting only `views` describes reach where the question is usually engagement, and these are precisely the queries a caller cannot widen themselves, since none of them accept `--metrics`.

**The deciding argument is newer than the issue.** Since #190 these commands *print* a notice saying `engagedViews` keeps the stricter definition across the 2026-08-24 change, while not returning it. Recommending a metric the command does not provide was the remaining incoherence.

## Why `sort` stays `-views`

Row order is part of a predefined report's shape. Adding a column is one break; silently reordering every row would be a second, with nothing forcing it. #179 settled the identical question the same way for custom queries: a query that ranked one way before the metric existed ranks identically after.

Verified rather than assumed. Row keys and `views` values are **identical before and after** on all four surfaces:

```
traffic-sources:      row order identical=True  views identical=True  n=8
geography:            row order identical=True  views identical=True  n=2
channel-search-terms: row order identical=True  views identical=True  n=19
```

## Breaking impact

Recorded in `BREAKING-CHANGES.md`. Measured before/after headers:

| command | before | after |
|---|---|---|
| `get-traffic-sources` | `insightTrafficSourceType, views` | `…, views, engagedViews` |
| `get-search-terms` | `insightTrafficSourceDetail, views` | `…, views, engagedViews` |
| `--report geography` | `country, views, estimatedMinutesWatched` | `country, views, engagedViews, estimatedMinutesWatched` |
| `get-channel-search-terms` | `…, views, estimatedMinutesWatched` | `…, views, engagedViews, estimatedMinutesWatched` |

`views` keeps its index everywhere. On the two reports that also carry `estimatedMinutesWatched`, **that column moves from index 2 to index 3**. Name-based readers are unaffected.

`--report demographics` is unchanged: `viewerPercentage` is a share, not a count.

## Incidental fixes

`get-traffic-sources` and `get-search-terms` read `row[1]` positionally. Both now resolve columns by name, matching what `get-channel-search-terms` already did, so the next column change cannot silently shift them.

Two `Output Fields` lists in `docs/commands/analytics.md` were wrong **before** this change, same class as #176: `get-search-terms` was documented as returning `estimated_minutes_watched`, which that query shape does not permit alongside `insightTrafficSourceDetail` filtered to `YT_SEARCH`, and both lists used invented snake_case names instead of the API's own. Corrected against measured output.

## E2E

Built binary against the live API:

- all four surfaces gained exactly `engagedViews`, existing columns in unchanged relative order
- `get-channel-search-terms --output pretty` reports `Total views from search: 198` / `Total engaged views from search: 121`, the 39% gap visible end to end
- `--report demographics` unchanged
- `--report bogus` still exits 1

Tests 246 -> 248. The #179 guard is inverted rather than deleted, plus two new guards pinning the sort axis and the column position. Type-check and lint clean.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -2,6 +2,50 @@
 
 ## Unreleased
 
+**The fixed-shape analytics reports gained `engagedViews`** (#185).
+
+Four surfaces send a metric set the caller cannot override with `--metrics`.
+All of them now carry `engagedViews`, inserted immediately after `views`:
+
+| command | before | after |
+|---|---|---|
+| `get-traffic-sources` | `insightTrafficSourceType, views` | `insightTrafficSourceType, views, engagedViews` |
+| `get-search-terms` | `insightTrafficSourceDetail, views` | `insightTrafficSourceDetail, views, engagedViews` |
+| `get-channel-analytics --report devices\|geography\|traffic-sources\|subscription-status` | `<dimension>, views, estimatedMinutesWatched` | `<dimension>, views, engagedViews, estimatedMinutesWatched` |
+| `get-channel-search-terms` | `insightTrafficSourceDetail, views, estimatedMinutesWatched` | `insightTrafficSourceDetail, views, engagedViews, estimatedMinutesWatched` |
+
+`--report demographics` is unchanged. It reports `viewerPercentage`, a share
+rather than a count, so no view metric exists there for the change to have
+moved.
+
+**Who is affected:** anything reading these outputs by column position.
+`views` keeps its index in all four, but on the two reports that also carry
+`estimatedMinutesWatched` that column **moves from index 2 to index 3**.
+Callers reading by column name are unaffected.
+
+**Row order is unchanged.** Every affected report keeps `sort: '-views'`, so
+rows rank exactly as before. Verified against the live API: row keys and
+`views` values are identical before and after for all four surfaces.
+
+**Why:** `views` and `engagedViews` are not interchangeable, and the gap is
+large. Measured channel-wide over 2026-05-01..2026-08-25:
+
+| report | `views` | `engagedViews` |
+|---|---:|---:|
+| geography | 1485 | 883 |
+| devices | 2849 | 1873 |
+| traffic-sources | 2873 | 1896 |
+| channel search terms | 189 | 115 |
+
+A breakdown reporting only `views` describes reach where the question is
+usually engagement. These four are also the queries a caller cannot fix
+themselves, since none of them accept `--metrics`.
+
+Since #190 these commands print a notice stating that `engagedViews` keeps the
+stricter definition across the 2026-08-24 counting change. Recommending a
+metric the command did not return was the remaining incoherence, and this
+closes it.
+
 **`get-video-analytics` default metrics gained `engagedViews`** (#175).
 
 The default set went from eight metrics to nine, with `engagedViews` inserted
--- a/commands/get-channel-search-terms.ts
+++ b/commands/get-channel-search-terms.ts
@@ -67,15 +67,17 @@ async function getChannelSearchTermsCommand(options: ChannelSearchTermsOptions):
     const colIndex = (name: string) =>
       columnHeaders.findIndex(h => h.name === name);
 
-    const idxTerm  = colIndex('insightTrafficSourceDetail');
-    const idxViews = colIndex('views');
-    const idxWatch = colIndex('estimatedMinutesWatched');
+    const idxTerm    = colIndex('insightTrafficSourceDetail');
+    const idxViews   = colIndex('views');
+    const idxEngaged = colIndex('engagedViews');
+    const idxWatch   = colIndex('estimatedMinutesWatched');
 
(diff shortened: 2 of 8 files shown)


prog893:
@coderabbitai full review

This PR was merged before you had reviewed it: your only comment here was the rate-limit notice, and there were no reviews on record. That was a process error on the merge side, not a signal that review was unwanted.

Requesting the full review now. Anything you find will be fixed in a follow-up PR against main rather than left standing.

Summary of what landed: `engagedViews` added to four fixed-shape analytics reports (`CHANNEL_REPORT_TYPES` devices/geography/traffic-sources/subscription-status, `fetchTrafficSources`, `fetchSearchTerms`, `SEARCH_TERMS_METRICS`), inserted immediately after `views` with `sort` deliberately left at `-views`. `get-traffic-sources` and `get-search-terms` also switched from positional `row[1]` reads to name-based column lookup. Two `Output Fields` doc lists were corrected. Breaking impact recorded in BREAKING-CHANGES.md: `estimatedMinutesWatched` shifts from index 2 to 3 on the two reports carrying it.

PR #200, opened 2026-08-24
feat(#185): add engagedViews to the fixed-shape analytics reports

### LINKED ISSUE #185: Analytics: decide whether the fixed report shapes should carry engagedViews

prog893:
Closes #185.

> Resubmitted. #199 was merged before CodeRabbit had reviewed it, so main was rolled back to `a6562de` and this PR reopened to go through the gate properly. The commit is unchanged (`e3b31c5`).

## The decision

**Widen all four surfaces.** `engagedViews` goes immediately after `views`; `sort` stays `-views`.

#185 asked for a per-surface decision rather than a global one. Having measured each, they all land the same way, for the same reason.

## Measured live, 2026-08-25

#179 deferred this because compatibility was unmeasured. Probed every shape directly:

| shape | `views,engagedViews` |
|---|---|
| `insightTrafficSourceType` filtered to a video | accepted, 9 rows |
| `insightTrafficSourceDetail`+`YT_SEARCH` filtered to a video | accepted |
| `insightTrafficSourceDetail`+`YT_SEARCH` channel-wide | accepted, 10 rows |
| `deviceType,operatingSystem` / `country` / `insightTrafficSourceType` / `subscribedStatus` | all accepted |
| `sort: -engagedViews` | also accepted |

The same run shows why it matters. Channel-wide over `2026-05-01..2026-08-25`:

| report | `views` | `engagedViews` | gap |
|---|---:|---:|---:|
| geography | 1485 | 883 | 41% |
| devices | 2849 | 1873 | 34% |
| traffic-sources | 2873 | 1896 | 34% |
| channel search terms | 189 | 115 | 39% |

Per-video long-form the two agree exactly (548 against 548). The channel-level gap is the Shorts mix. So a breakdown reporting only `views` describes reach where the question is usually engagement, and these are precisely the queries a caller cannot widen themselves, since none of them accept `--metrics`.

**The deciding argument is newer than the issue.** Since #190 these commands *print* a notice saying `engagedViews` keeps the stricter definition across the 2026-08-24 change, while not returning it. Recommending a metric the command does not provide was the remaining incoherence.

## Why `sort` stays `-views`

Row order is part of a predefined report's shape. Adding a column is one break; silently reordering every row would be a second, with nothing forcing it. #179 settled the identical question the same way for custom queries: a query that ranked one way before the metric existed ranks identically after.

Verified rather than assumed. Row keys and `views` values are **identical before and after** on all four surfaces:

```
traffic-sources:      row order identical=True  views identical=True  n=8
geography:            row order identical=True  views identical=True  n=2
channel-search-terms: row order identical=True  views identical=True  n=19
```

## Breaking impact

Recorded in `BREAKING-CHANGES.md`. Measured before/after headers:

| command | before | after |
|---|---|---|
| `get-traffic-sources` | `insightTrafficSourceType, views` | `…, views, engagedViews` |
| `get-search-terms` | `insightTrafficSourceDetail, views` | `…, views, engagedViews` |
| `--report geography` | `country, views, estimatedMinutesWatched` | `country, views, engagedViews, estimatedMinutesWatched` |
| `get-channel-search-terms` | `…, views, estimatedMinutesWatched` | `…, views, engagedViews, estimatedMinutesWatched` |

`views` keeps its index everywhere. On the two reports that also carry `estimatedMinutesWatched`, **that column moves from index 2 to index 3**. Name-based readers are unaffected.

`--report demographics` is unchanged: `viewerPercentage` is a share, not a count.

## Incidental fixes

`get-traffic-sources` and `get-search-terms` read `row[1]` positionally. Both now resolve columns by name, matching what `get-channel-search-terms` already did, so the next column change cannot silently shift them.

Two `Output Fields` lists in `docs/commands/analytics.md` were wrong **before** this change, same class as #176: `get-search-terms` was documented as returning `estimated_minutes_watched`, which that query shape does not permit alongside `insightTrafficSourceDetail` filtered to `YT_SEARCH`, and both lists used invented snake_case names instead of the API's own. Corrected against measured output.

## E2E

Built binary against the live API:

- all four surfaces gained exactly `engagedViews`, existing columns in unchanged relative order
- `get-channel-search-terms --output pretty` reports `Total views from search: 198` / `Total engaged views from search: 121`, the 39% gap visible end to end
- `--report demographics` unchanged
- `--report bogus` still exits 1

Tests 246 -> 248. The #179 guard is inverted rather than deleted, plus two new guards pinning the sort axis and the column position. Type-check and lint clean.

## Summary by CodeRabbit

- **New Features**
  - Analytics reports now include an `engagedViews` metric alongside `views` for search terms, traffic sources, channel search terms, and applicable channel reports.
  - Pretty-printed reports display engaged-view counts per row and in totals.
  - Results remain ranked by views, with demographics reporting unchanged.

- **Documentation**
  - Updated command, tool, and analytics documentation to describe the new metric, column ordering, and view-counting notice.

- **Breaking Changes**
  - Some report columns shift position because `engagedViews` is inserted after `views`.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -2,6 +2,50 @@
 
 ## Unreleased
 
+**The fixed-shape analytics reports gained `engagedViews`** (#185).
+
+Four surfaces send a metric set the caller cannot override with `--metrics`.
+All of them now carry `engagedViews`, inserted immediately after `views`:
+
+| command | before | after |
+|---|---|---|
+| `get-traffic-sources` | `insightTrafficSourceType, views` | `insightTrafficSourceType, views, engagedViews` |
+| `get-search-terms` | `insightTrafficSourceDetail, views` | `insightTrafficSourceDetail, views, engagedViews` |
+| `get-channel-analytics --report devices\|geography\|traffic-sources\|subscription-status` | `<dimension>, views, estimatedMinutesWatched` | `<dimension>, views, engagedViews, estimatedMinutesWatched` |
+| `get-channel-search-terms` | `insightTrafficSourceDetail, views, estimatedMinutesWatched` | `insightTrafficSourceDetail, views, engagedViews, estimatedMinutesWatched` |
+
+`--report demographics` is unchanged. It reports `viewerPercentage`, a share
+rather than a count, so no view metric exists there for the change to have
+moved.
+
+**Who is affected:** anything reading these outputs by column position.
+`views` keeps its index in all four, but on the two reports that also carry
+`estimatedMinutesWatched` that column **moves from index 2 to index 3**.
+Callers reading by column name are unaffected.
+
+**Row order is unchanged.** Every affected report keeps `sort: '-views'`, so
+rows rank exactly as before. Verified against the live API: row keys and
+`views` values are identical before and after for all four surfaces.
+
+**Why:** `views` and `engagedViews` are not interchangeable, and the gap is
+large. Measured channel-wide over 2026-05-01..2026-08-25:
+
+| report | `views` | `engagedViews` |
+|---|---:|---:|
+| geography | 1485 | 883 |
+| devices | 2849 | 1873 |
+| traffic-sources | 2873 | 1896 |
+| channel search terms | 189 | 115 |
+
+A breakdown reporting only `views` describes reach where the question is
+usually engagement. These four are also the queries a caller cannot fix
+themselves, since none of them accept `--metrics`.
+
+Since #190 these commands print a notice stating that `engagedViews` keeps the
+stricter definition across the 2026-08-24 counting change. Recommending a
+metric the command did not return was the remaining incoherence, and this
+closes it.
+
 **`get-video-analytics` default metrics gained `engagedViews`** (#175).
 
 The default set went from eight metrics to nine, with `engagedViews` inserted
--- a/commands/get-channel-search-terms.ts
+++ b/commands/get-channel-search-terms.ts
@@ -67,15 +67,17 @@ async function getChannelSearchTermsCommand(options: ChannelSearchTermsOptions):
     const colIndex = (name: string) =>
       columnHeaders.findIndex(h => h.name === name);
 
-    const idxTerm  = colIndex('insightTrafficSourceDetail');
-    const idxViews = colIndex('views');
-    const idxWatch = colIndex('estimatedMinutesWatched');
+    const idxTerm    = colIndex('insightTrafficSourceDetail');
+    const idxViews   = colIndex('views');
+    const idxEngaged = colIndex('engagedViews');
+    const idxWatch   = colIndex('estimatedMinutesWatched');
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const structuredRows = rows.map((row: any[]) => ({
       rank: 0,           // filled below
       searchTerm:        row[idxTerm]  as string,
       views:             row[idxViews] as number,
+      engagedViews:      idxEngaged >= 0 ? row[idxEngaged] as number : 0,
       watchTimeMinutes:  idxWatch >= 0 ? row[idxWatch] as number : 0,
     }));
 
@@ -153,7 +155,8 @@ async function getChannelSearchTermsCommand(options: ChannelSearchTermsOptions):
         console.log('');
 
         let totalViews = 0;
-        structuredRows.forEach(r => { totalViews += r.views; });
+        let totalEngagedViews = 0;
+        structuredRows.forEach(r => { totalViews += r.views; totalEngagedViews += r.engagedViews; });
 
         structuredRows.forEach(r => {
           const pct = totalViews > 0 ? ((r.views / totalViews) * 100).toFixed(1) : '0.0';
@@ -163,6 +166,9 @@ async function getChannelSearchTermsCommand(options: ChannelSearchTermsOptions):
             chalk.gray('      Views:      ') + chalk.cyan(formatNumber(r.views)) +
             chalk.gray(` (${pct}% of search traffic)`)
           );
+          console.log(
+            chalk.gray('      Engaged:    ') + chalk.cyan(formatNumber(r.engagedViews))
+          );
           if (r.watchTimeMinutes > 0) {
             const watchHours = (r.watchTimeMinutes / 60).toFixed(0);
             console.log(chalk.gray('      Watch time:  ') + chalk.cyan(`${formatNumber(parseInt(watchHours, 10))}h`));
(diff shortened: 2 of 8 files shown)


prog893:
@coderabbitai Six of seven applied in `af61929`, one accepted on different grounds than stated.

| # | finding | outcome |
|---|---|---|
| 1 | `engagedViews` missing from `--output text` on both search-term commands | **fixed** |
| 2 | scope the `pretty` case with braces | **applied**, premise corrected below |
| 3 | top-N sums labelled as complete search totals | **fixed** |
| 4 | MCP descriptions use measurements from the wrong scope | **fixed** |
| 5 | column contract not scoped to predefined reports | **fixed** |
| 6 | unlabelled fence (MD040) | **fixed** |
| 7 | stale comments describing views-only | **fixed** |

**3 is the one that mattered.** Both reports cap their rows, `--limit` on the per-video one and `CHANNEL_SEARCH_TERMS_MAX_RESULTS` on the channel one, so "Total views from search" claimed more than was measured. Worth noting this was already wrong for `views` before this PR; adding `engagedViews` under the same label would have doubled an existing error rather than introducing a new one. Both lines are now scoped to the rows actually shown: `Views from these 20 terms: 212`.

**1 was a real gap.** The metric was mapped into the row objects and rendered in `pretty`, but the tab-separated writers still emitted the old column list, so the format most likely to be piped was the one missing the new column.

**4 was a copy-paste error worth catching.** The same note went into all three tool descriptions, so the video-scoped `youtube_get_search_terms` cited the channel-wide traffic-source figure. Each tool now carries a measurement from its own scope: 548/548 and 197/8 for the per-video breakdowns, 189/115 for channel search terms, 2873/1896 for channel traffic sources.

**On 2, the premise does not hold, though the change is still worth making.** The finding attributes the rule to Biome and states the change is needed so `bun run lint` passes. This repo has no Biome: the toolchain is ESLint via `eslint.config.mjs`, there is no `biome.json`, and `bun run lint` reports `0 errors` both before and after (4 pre-existing `no-explicit-any` warnings, none in this file).

I applied it anyway on the underlying merit: switch clauses share a single lexical scope, so a `let` in an unbraced case is a genuine footgun whether or not the configured linter flags it. Flagging the tool attribution only so the claim about `bun run lint` failing does not get treated as verified.

Verified end to end against the live API after the fixes: text output emits `rank/searchTerm/views/engagedViews/watchTimeMinutes`, and pretty totals read `Views from these 20 terms: 212` / `Engaged views from these 20 terms: 129`.

Type-check clean, lint 0 errors, 248 tests pass. Please re-review and state an explicit verdict.

prog893:
@coderabbitai Fixed in `d16678f`, and the finding was broader than stated.

You flagged `youtube_get_traffic_sources`. The same caveat you attached to it ("the existing 548/548 and 197/8 examples are suitable only if they were measured from that tool's query shape") also condemned my fix to `youtube_get_search_terms`: I had given it 548/548, which came from the **traffic-source** query, not the search-term one. So the previous commit corrected the scope for channel search terms and left the identical error in both video-scoped tools.

Rather than reuse a number, I measured each tool's own query shape against the live API over `2026-05-01..2026-08-25`:

**`insightTrafficSourceType` filtered to one video** (`youtube_get_traffic_sources`)

| video | `views` | `engagedViews` |
|---|---:|---:|
| long-form | 553 | 553 |
| long-form | 241 | 241 |
| Short | 197 | 8 |
| Short | 245 | 23 |

**`insightTrafficSourceDetail` + `YT_SEARCH` filtered to one video** (`youtube_get_search_terms`)

| video | `views` | `engagedViews` |
|---|---:|---:|
| long-form | 53 | 53 |
| long-form | 2 | 2 |
| Short | 131 | 62 |
| Short | 2 | 0 |

Both descriptions now cite their own shape, and the framing changed with the data: the useful signal for a caller is not a single ratio but the long-form/Shorts split. The metrics agree exactly on long-form and diverge sharply on Shorts, and neither tool can tell which kind of video it was pointed at.

`youtube_get_channel_search_terms` keeps 189/115, which was already measured channel-wide from its own shape.

Every numeric example in these three descriptions is now measured from the query the tool actually issues. Type-check clean, lint 0 errors, 248 tests pass.

On the sandbox note: understood that you could not run the checks yourself. The commands are `bun run type-check`, `bun run lint`, `bun test`, and the live measurements above came from `youtubeAnalytics.reports.query` against the authenticated channel.

Please re-review and state an explicit verdict.

Issue #198, opened 2026-08-24
Error handling D: ban bare catch {} by lint, after the three genre fixes land


prog893:
Split out of #194 by genre. Enforcement, so the class does not regrow once #195, #196 and #197 land.

## Current state

11 catch blocks have no statement in the body at all:

```
lib/auth.ts:160          lib/completion.ts:169     lib/completion.ts:192
lib/cache.ts:453         lib/cache.ts:459          lib/reports.ts:312
lib/version.ts:26        lib/youtube.ts:216        commands/cache.ts:57
commands/cache.ts:66     commands/config.ts:25
```

A bare `catch {}` records nothing about what was expected, so the next reader cannot tell a deliberate decision from an oversight. Every one of the eleven is claimed by one of the three genre issues, which is the argument for the rule: after those land there should be no reason to write another.

## Proposed rule

```jsonc
"@typescript-eslint/no-empty": ["error", { "allowEmptyCatch": false }]
```

This catches the mechanical case only. A catch body that logs nothing but returns a value (`return null`, `return {}`) still passes, and those are #195's territory, not a lint's.

The rule to write down alongside it, since lint cannot express it:

> Swallow an error only when it is retryable or an expected control-flow outcome. In the second case, name the error you expect (`err.code === 'ENOENT'`) and re-throw or `debug()` anything else.

`lib/lock.ts:125` is the exemplar to point at:

```ts
} catch (err) {
  const error = err as NodeJS.ErrnoException;
  // ESRCH = no such process
  return error.code !== 'ESRCH';
}
```

## Sequencing

**Blocked on #195, #196 and #197.** Enabling the rule first turns CI red on eleven pre-existing sites and forces either a rushed fix or eleven suppression comments, which is worse than the current state. Land the three genre fixes, then enable the rule as the final commit with no violations left to suppress.

Also worth adding to `docs/development/error-handling.md`, which currently covers user-facing error messages but says nothing about when swallowing is acceptable.

Issue #197, opened 2026-08-24
Error handling C: fallback probes hide real failures, incl. a network outage misreported as a credential failure


prog893:
Split out of #194 by genre. Subtler than #195 and #196: the code is structured as "try this, and if it does not work do something else", so a genuine failure of the first branch is hard to tell apart from that branch simply not applying.

> **Corrected after verification against `f7053e3`.** The original framing of `lib/youtube.ts:216` and `lib/completion.ts:169` did not survive measurement, and one site was missing. Evidence in the [verification comment](https://github.com/prog893/staqan-yt-cli/issues/197#issuecomment-5438832986). This body reflects the corrected findings.

## Sites

| site | probe | intended miss | actually indistinguishable from | verdict |
|---|---|---|---|---|
| `lib/auth.ts:133` | `refreshAccessToken()` | revoked refresh token | **network outage, Google 5xx** | wrong answer, highest cost |
| `lib/youtube.ts:216` | `channels.list` by ID | (none: the miss never throws) | every genuine failure | dead catch, wasted quota |
| `lib/version.ts:26` | `require(package.json)` per candidate layout | wrong layout for this install | a syntax error in the correct `package.json` | confirmed, low severity |
| `lib/completion.ts:169` | `execSync('ps -p $$ -o comm=')` | `ps` unavailable | n/a, the probe cannot work at all | separate defect |
| `lib/utils.ts:849` | `Intl.DateTimeFormat().resolvedOptions().timeZone` | environment has no ICU data | any other `Intl` failure | confirmed, inert |
| `lib/auth.ts:160` | `new URL(uri)` over a redirect-URI list | that entry is malformed, skip it | nothing much, this one is fine | no change |

## The one that matters: `lib/auth.ts:133`

```ts
} catch {
  throw new Error('Failed to refresh token. Please re-authenticate: staqan-yt auth');
}
```

A bare catch, so a dropped connection, a Google 5xx and a genuinely revoked `refresh_token` all collapse into one message prescribing an OAuth re-run. Measured with connectivity blocked identically in two runs, varying only whether the token had expired:

| token state | message |
|---|---|
| expired, refresh attempted | `✗ Failed to refresh token. Please re-authenticate: staqan-yt auth` |
| fresh, refresh skipped | `✗ Unable to connect. Is the computer able to access the url?` |

The same network condition is reported accurately when this catch is bypassed, and as a credential problem when it is not. Acting on that advice discards a working session in response to a network blip.

[...]

Issue #196, opened 2026-08-24
Error handling B: best-effort cleanup catches swallow more than the errno they document (13 sites)


prog893:
Split out of #194 by genre. Lowest severity of the four, and the cheapest to fix.

> **Corrected 2026-08-27 after verification against `f7053e3`.** The original write-up listed 8 sites and attributed three of them to `mkdir`/`EEXIST`. Verification found the count low and those three rows wrong. Details and the end-to-end reproduction are in the [verification comment](https://github.com/prog893/staqan-yt-cli/issues/196#issuecomment-5438806379); this body now reflects the corrected findings.

## The pattern

Best-effort filesystem operations whose failure genuinely does not matter, guarded by a catch that swallows everything rather than the one error that is expected.

```ts
try {
  await fs.unlink(csvPath);
} catch {
  // Ignore if file doesn't exist
}
```

The comment says `ENOENT`. The code says everything.

## Sites that discard the error outright (13)

Ranked by whether anything observable follows.

| site | operation | expected error | also swallows | impact |
|---|---|---|---|---|
| `commands/cache.ts:66` | `unlink` cache file, gates `removed++` | `ENOENT` | `EACCES`, `EPERM`, `EBUSY`, `EROFS` | **wrong count, wrong message** |
| `commands/cache.ts:57` | `readdir` cache dir | `ENOENT` | `EACCES`, `ENOTDIR` | per-channel targets silently skipped |
| `commands/config.ts:22` | `unlink` per-channel completion cache | `ENOENT` | same as above | stale completions for the old channel |
| `commands/config.ts:25` | `readdir` cache dir | `ENOENT` | `EACCES`, `ENOTDIR` | invalidation silently does nothing |
| `lib/cache.ts:453` | `unlink` cached CSV | `ENOENT` | `EACCES`, `EIO`, `EBUSY`, `EPERM` | orphaned cache file, index says deleted |
| `lib/cache.ts:459` | `unlink` sidecar metadata | `ENOENT` | same | same |
| `lib/reports.ts:157` | `unlink` partial download, HTTP 429 path | n/a | everything | leaked temp file |
| `lib/reports.ts:167` | `unlink` partial download, non-200 path | n/a | everything | leaked temp file |
| `lib/reports.ts:187` | `unlink` partial download, `pipeline` failure | n/a | everything | leaked temp file |
| `lib/reports.ts:203` | `unlink` partial download, socket error | n/a | everything | leaked temp file |
[...]

Issue #192, opened 2026-08-24
Cache: a corrupt sidecar is indistinguishable from a missing one and silently disables CSV validation


prog893:
## Problem

`loadReportMetadata` (`lib/cache.ts:273`) swallows every error and returns `null`:

```ts
try {
  const data = await fs.readFile(metadataPath, 'utf-8');
  return JSON.parse(data) as ReportMetadata;
} catch {
  return null;
}
```

A truncated or malformed sidecar is therefore indistinguishable from one that was never written, and both degrade silently at all three call sites. The worst is `readCachedReport` (`lib/cache.ts:384`):

```ts
const metadata = await loadReportMetadata(...);
if (metadata) {
  if (parsed.headers.join(',') !== metadata.columns.join(',')) return null;   // skipped
  if (!metadata.isComplete) return null;                                      // skipped
}
return parsed;
```

When the sidecar is unreadable, **both integrity checks are skipped and the CSV is served unvalidated**. The one mechanism protecting against a corrupted cached report is disabled precisely by a second corruption, with no output of any kind. `getActualDates` (`:608`) similarly falls back to the API window, silently widening that report's apparent coverage.

## Proposed behaviour

Rebuild the sidecar from what is actually known, and never fabricate what is not.

**From the index entry** (authoritative, already in memory during any query):
`reportId`, `reportTypeId`, `channelId`, `startTime`, `endTime`, `createTime`, `downloadedAt`, `expiresAt`, `fileSize`, `row_count`

**Re-measured from the CSV on disk** (measured, not inferred):
`columns` (its header row), `startTimeActual` / `endTimeActual` (min and max of its `date` column)

**Not recoverable, and must not be invented:**
`jobId`, `downloadUrl`, `isComplete`

The last group is the constraint that shapes the design. A rebuilt sidecar cannot claim `isComplete: true`, because nothing on disk establishes it. It cannot synthesise a `downloadUrl` or `jobId`, because those came from an API response that is gone. Writing a plausible-looking value for any of them would convert a detectable gap into an undetectable lie, which is worse than the current silence.

That implies:

1. `ReportMetadata` grows a marker distinguishing a rebuilt sidecar from an original, so the unrecoverable fields are readably absent rather than defaulted.
2. `jobId` / `downloadUrl` / `isComplete` become optional, or the rebuilt record uses a narrower type.
[...]

PR #201, opened 2026-08-27
fix(#195): tell a damaged optional-JSON file apart from an absent one

### LINKED ISSUE #195: Error handling A: optional-JSON loaders report a corrupt file as absent (8 sites)
Split out of #194 by genre. This is the highest-impact of the four.

## The pattern

Eight sites repeat the same shape: read an optional file, parse it, and return a neutral value on any failure.

> Corrected from six after a full sweep of `lib/`, `commands/` and `bin/`. `lib/config.ts:67` and `lib/cache.ts:55` were missed in the original count. Full inventory, including the sites that correctly stay as they are, is in the scope-audit comment below.

```ts
try {
  const content = await fs.readFile(PATH, 'utf-8');
  return JSON.parse(content) as T;
} catch {
  return null;
}
```

That conflates two different situations:

- **absent** (`ENOENT`): expected. Returning `null` is correct control flow.
- **damaged** (JSON syntax error, `EACCES`, `EIO`): unexpected. Returning `null` reports the file as "not there" when it is there and broken.

## Sites

| site | what it loads | what a corrupt file looks like |
|---|---|---|
| `lib/auth.ts:30` | `credentials.json` | "not authenticated", sending the reader to `staqan-yt auth` for a file that already exists |
| `lib/auth.ts:42` | `token.json` | a silent full re-auth instead of naming the damaged file |
| `lib/cache.ts:283` | report metadata sidecar | CSV column and `isComplete` validation silently skipped (#192) |
| `lib/config.ts:53` (`loadRawConfig`) | `config.json` | every setting reverts to defaults with no message |
| `lib/config.ts:67` (`loadConfig`) | `config.json` | same, and this is the one every command actually calls; `loadRawConfig` only serves `config list` and `config get` |
| `lib/cache.ts:55` (`loadCacheIndex`) | `cache-index.json` | the archive files stay on disk but become invisible, so the next fetch re-downloads work already held locally and then writes an index that no longer references the old files |
| `lib/youtube.ts:17` | handle cache | reads as an empty cache, so every handle is re-resolved against the API |
| `commands/complete.ts:34` | completion cache | reads as an empty cache |

`config.json` is the sharpest after the sidecar. A stray comma from a hand-edit makes `default.channel` and `default.output` silently stop applying, and the CLI behaves exactly as though they had never been set. There is no output at any verbosity that would point at the file.

`lib/auth.ts` is a close second, because its failure message actively misdirects: it names the fix for a different problem.

## Proposed work

One shared helper, replacing all eight:

```ts
/** Parse a JSON file, or null if it does not exist. Throws if it exists and cannot be read or parsed. */
async function loadJsonIfPresent<T>(path: string): Promise<T | null>
```

- `ENOENT` returns `null`, which is the only case any caller currently means by `null`.
- Anything else throws, carrying the path and the underlying error.

Callers then decide per site whether to fail loudly (`auth`, `config`) or warn and continue with a cold cache (`youtube`, `complete`, `cache`). The point is that each site makes that decision explicitly rather than inheriting silence from the loader. For the cache sites the warning must fire on damage only: an absent index or an absent completion cache is the ordinary first-run state and has to stay silent.

`lib/lock.ts:125` is the pattern to copy. It is the one place in the codebase that already does this correctly:

```ts
} catch (err) {
  const error = err as NodeJS.ErrnoException;
  // ESRCH = no such process
  return error.code !== 'ESRCH';
}
```

It names the error it expects and treats everything else as a different answer. Note that this is `isProcessAlive`, not the lock-file read at `lib/lock.ts:91`, which has the anti-pattern shape but is out of scope: both of its failure modes have the same correct answer, so the swallow there is deliberate. Reasoning in the scope-audit comment.

## Scope note

This touches `lib/auth.ts` and `lib/config.ts`, which every command depends on, so it warrants its own PR with real coverage rather than being folded into the cache work. #192 depends on this landing first: rebuilding a corrupt sidecar is not reachable until `loadReportMetadata` can tell absent from damaged.

prog893:
Closes #195.

## The problem

Eight loaders read an optional JSON file and returned a neutral value on *any* failure, collapsing two different situations into one answer:

- **absent** (`ENOENT`): expected, and `null` is the correct control flow.
- **damaged** (JSON syntax error, `EACCES`, `EIO`): the file is there and unusable.

Because every caller reads `null` as "not configured yet" and prints guidance to match, a damaged file sent the reader to fix the wrong problem. `credentials.json` produced "run the auth command first" for a file that already existed. A stray comma in `config.json` made `default.channel` and `default.output` silently stop applying, with no output at any verbosity pointing at the file.

## The change

One helper, `loadJsonIfPresent<T>(path, label)` in `lib/utils.ts`, replacing all eight call sites. `ENOENT` returns `null`, which is the only case any caller ever meant by `null`. Everything else throws, carrying the path, the underlying parse error, and the remedy.

Each site then decides explicitly what to do, rather than inheriting silence from the loader:

| site | absent | damaged |
|---|---|---|
| `lib/auth.ts` credentials, token | setup guidance | throws, naming the file |
| `lib/config.ts` `loadConfig`, `loadRawConfig` | defaults | throws, naming the file |
| `lib/cache.ts` cache index | silent, fresh index | warns, fresh index |
| `lib/cache.ts` report sidecar | `null` | warns, `null` |
| `lib/youtube.ts` handle cache | silent, cold | warns, cold |
| `commands/complete.ts` completion cache | silent, cold | `debug()` only, cold |

The split follows what the caller can actually do about it. Config and credentials decide what a command does at all, so a damaged one stops the run. A cache is optional by definition and must not abort the command, but staying silent turns a one-off corruption into a permanent extra API call per handle with nothing to point at, so it warns and continues.

The completion cache is the one site that must stay quiet on **both** streams no matter what: it runs inside the interactive shell on every Tab, where a warning would print into the middle of the prompt. The error goes to `debug()`, so `-v` can still explain a cache that never seems to hit.

`loadCacheIndex` also gained a case it did not have before. A file that parses but is not index-shaped is the same class as a parse failure (present, unusable), so it is now reported the same way instead of falling into the silent `debug()` path.

## E2E proof

Built `main` (`f7053e3`) and this branch side by side, and ran identical commands against fake `HOME` / `STAQAN_YT_DATA_DIR` trees. No real credentials or archive touched.

**Corrupt `config.json`** (`{ "default": { "channel": "@staqan", }, }`):

```
before   exit=0   default.channel: (not set) / default.output: pretty
after    exit=1   ✗ config (…/config.json) is not valid JSON: JSON Parse error:
                    Property name must be a string literal. Fix the file, or delete it to start over.
```

On `main` that output is **byte-identical to a cold start**, which is precisely the defect.

**Corrupt `credentials.json`:**

```
before   ✗ Credentials not found. Please run the auth command first.
after    ✗ OAuth credentials (…/credentials.json) is not valid JSON: JSON Parse error:
           Unexpected identifier "not". Fix the file, or delete it to start over.
```

**Cache index**, with an archive CSV still on disk:

| index state | before | after |
|---|---|---|
| damaged JSON | silent, archive invisible | warns, names the file and the rebuild command |
| parses, wrong shape | silent, archive invisible | warns, names the file |
| version mismatch | warns | **identical to before** |
| absent | silent | **identical to before** |

**Handle cache** and **report sidecar**: damaged now warns and continues; absent is unchanged and silent; a valid sidecar still parses.

**Completion cache** (the non-regression that matters most), measured by byte count per stream:

```
completionCache=damaged  before  exit=0 stdout=0 stderr=0
completionCache=damaged  after   exit=0 stdout=0 stderr=0
completionCache=absent   before  exit=0 stdout=0 stderr=0
completionCache=absent   after   exit=0 stdout=0 stderr=0
```

**Controls.** Cold start stays silent on every path, and a valid config still applies: `config get default.channel` returns `@staqan` and `config list` still honours `output=csv`, identical on both builds.

## Checks

`type-check` clean, `lint` 0 errors (4 pre-existing `any` warnings in `commands/mcp.ts`), **258 tests pass**. Coverage added in `tests/json-loader.test.ts` (the absent/damaged/permission split) and `tests/cache-store.test.ts` (index damage vs. cold start, so the silent-first-run case cannot regress).

## Follow-ups

#192 depends on this landing: rebuilding a damaged sidecar is not reachable until `loadReportMetadata` can tell absent from damaged. #198 (lint rule banning bare `catch {}`) is the last of the family, after #196 and #197.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of missing, unreadable, and malformed configuration, credential, token, cache, and metadata files.
  - Added clearer warnings and recovery guidance when stored data is damaged or has an unexpected format.
  - Prevented corrupted files from being silently treated as missing or reset automatically.
- **Tests**
  - Added coverage for valid, missing, empty, malformed, and unreadable JSON files, including cache index recovery behavior.

### DIFF
--- a/commands/complete.ts
+++ b/commands/complete.ts
@@ -12,7 +12,7 @@ import { getChannelVideos, listChannelPlaylists, getChannelId } from '../lib/you
 import { getConfigValue } from '../lib/config';
 import { getAuthenticatedClient } from '../lib/auth';
 import { acquireLock, getLockPath } from '../lib/lock';
-import { CACHE_DIR, debug } from '../lib/utils';
+import { CACHE_DIR, debug, loadJsonIfPresent } from '../lib/utils';
 import { CompletionType, CompletionCache } from '../types';
 
 const TTL: Record<CompletionType, number> = {
@@ -29,9 +29,15 @@ function getChannelCachePath(channelId: string): string {
 }
 
 async function loadCache(cachePath: string): Promise<CompletionCache> {
+  // Completion runs inside the user's interactive shell on every Tab, so this
+  // is the one site that must stay quiet on stdout and stderr no matter what:
+  // a warning here would print into the middle of the prompt. The error still
+  // goes to debug(), so `-v` can explain a cache that never seems to hit
+  // (#195).
   try {
-    return JSON.parse(await fs.readFile(cachePath, 'utf-8'));
-  } catch {
+    return (await loadJsonIfPresent<CompletionCache>(cachePath, 'completion cache')) || {};
+  } catch (err) {
+    debug(`Ignoring unreadable completion cache: ${(err as Error).message}`);
     return {};
   }
 }
--- a/commands/complete.ts
+++ b/commands/complete.ts
@@ -29,11 +29,10 @@ function getChannelCachePath(channelId: string): string {
 }
 
 async function loadCache(cachePath: string): Promise<CompletionCache> {
-  // Completion runs inside the user's interactive shell on every Tab, so this
-  // is the one site that must stay quiet on stdout and stderr no matter what:
-  // a warning here would print into the middle of the prompt. The error still
-  // goes to debug(), so `-v` can explain a cache that never seems to hit
-  // (#195).
+  // Runs in the interactive shell on every Tab, so this must stay quiet on
+  // stdout and stderr: a warning would print into the middle of the prompt.
+  // The error goes to debug(), so `-v` can still explain a cache that never
+  // seems to hit.
   try {
     return (await loadJsonIfPresent<CompletionCache>(cachePath, 'completion cache')) || {};
   } catch (err) {
--- a/lib/auth.ts
+++ b/lib/auth.ts
@@ -22,22 +22,20 @@ const SCOPES = [
 ];
 
 /**
- * Load OAuth2 credentials from file
+ * Load OAuth2 credentials from file.
+ * Null if absent, so the caller prints setup guidance. Throws if damaged,
+ * since re-running auth does not repair a malformed credentials.json.
  */
 async function loadCredentials(): Promise<OAuth2Credentials | null> {
-  // Absent means "not set up yet" and the caller prints setup guidance.
-  // Damaged throws, because that guidance is wrong for a file that exists:
-  // re-running auth does not repair a malformed credentials.json (#195).
   return loadJsonIfPresent<OAuth2Credentials>(CREDENTIALS_PATH, 'OAuth credentials');
 }
 
 /**
- * Load saved token from file
+ * Load saved token from file.
+ * Null if absent, which triggers the sign-in prompt. Throws if damaged, so a
+ * corrupt token is not answered with a silent full re-auth.
  */
 async function loadToken(): Promise<OAuth2Token | null> {
-  // Same split as loadCredentials: absent triggers the sign-in prompt, damaged
-  // is reported so a corrupt token is not silently mistaken for a missing one
-  // and answered with a full re-auth (#195).
   return loadJsonIfPresent<OAuth2Token>(TOKEN_PATH, 'auth token');
 }
 
--- a/lib/auth.ts
+++ b/lib/auth.ts
@@ -10,6 +10,7 @@ import {
   TOKEN_PATH,
   ensureConfigDir,
   info,
+  loadJsonIfPresent,
 } from './utils';
 import { OAuth2Credentials, OAuth2Token } from '../types';
 
@@ -24,24 +25,20 @@ const SCOPES = [
  * Load OAuth2 credentials from file
  */
 async function loadCredentials(): Promise<OAuth2Credentials | null> {
-  try {
-    const content = await fs.readFile(CREDENTIALS_PATH, 'utf-8');
-    return JSON.parse(content) as OAuth2Credentials;
-  } catch {
-    return null;
-  }
+  // Absent means "not set up yet" and the caller prints setup guidance.
+  // Damaged throws, because that guidance is wrong for a file that exists:
+  // re-running auth does not repair a malformed credentials.json (#195).
+  return loadJsonIfPresent<OAuth2Credentials>(CREDENTIALS_PATH, 'OAuth credentials');
 }
 
 /**
  * Load saved token from file
  */
 async function loadToken(): Promise<OAuth2Token | null> {
-  try {
-    const content = await fs.readFile(TOKEN_PATH, 'utf-8');
-    return JSON.parse(content) as OAuth2Token;
-  } catch {
-    return null;
-  }
+  // Same split as loadCredentials: absent triggers the sign-in prompt, damaged
+  // is reported so a corrupt token is not silently mistaken for a missing one
+  // and answered with a full re-auth (#195).
+  return loadJsonIfPresent<OAuth2Token>(TOKEN_PATH, 'auth token');
 }
 
 /**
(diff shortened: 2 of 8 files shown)


prog893:
@coderabbitai Round 1 reconciled at `7ffda40`. Two findings fixed, three contested with measurements.

## Fixed

**`lib/cache.ts:90`, validate each `CacheIndexEntry`.** Correct, and reproduced against the built CLI before fixing. An index of `{"version":"2.0","lastUpdated":"...","entries":[null]}` passed the envelope check and reached a consumer:

```
before:  loadCacheIndex returned entries=[null]
         findCachedReports THREW: null is not an object (evaluating 'entry.channelId')
after:   ⚠ Cache index has an unexpected structure, treating the archive as empty.
         ⚠   To rebuild: staqan-yt fetch-reports --channel @staqan
         loadCacheIndex returned entries=[]
         findCachedReports ok, n=0
```

A damaged index surfacing as a property access on `null`, from a call site that never names the index, is exactly the failure this PR exists to remove. Added `isCacheIndexEntry`, checking the required fields only, since `createTime` and `row_count` are optional by design. Three tests: null member, partial member, and a complete entry that must still be accepted so the guard cannot over-reject.

**`lib/utils.ts:57`, recovery guidance on non-ENOENT read errors.** Correct and inconsistent as written: the parse branch two lines below said how to recover and the read branch did not. Now appends the readable-file and permissions check, with a test pinning it. The assertion targets our own wording rather than the errno text, which differs between macOS and Linux CI.

258 tests to 262.

## Contested

**`lib/utils.ts:49` and `lib/cache.ts:99`, "Everything is a named flag. No positional arguments at all."**

That rule is quoted out of its scope. In `CLAUDE.md` it is the heading of section 3, **AWS API Naming Conventions**, and every example under it is a shell invocation:

```
get-video --video-id <id>          # Get ONE video
get-videos --video-ids <id1> <id2> # Get MULTIPLE videos (batch)
```

It governs the CLI surface, where the project rejects `get-video <id>` in favour of `--video-id <id>`. It is not a TypeScript signature convention, and reading it as one contradicts the codebase:

- **36** multi-argument positional function declarations in `lib/` alone, including `getReportPaths(channelId, reportId, reportTypeId)` and `loadReportMetadata(channelId, reportId, reportTypeId)`, the direct neighbours of the code under review.
- **0** functions in `lib/` take an options object as their only parameter. There is no existing instance of the suggested pattern to be consistent with.

The `lib/cache.ts:99` comment also asks to convert `path.join` and `fs` calls to named option objects. Those are Node standard library signatures and cannot be changed, which is a reasonable signal that the rule is being applied outside its domain.

Converting `loadJsonIfPresent` alone would make it the only function in `lib/` with that shape, so the change would reduce consistency rather than increase it. Declining both.

**`tests/json-loader.test.ts:30`, "Always use try/finally to ensure cleanup."**

Same issue. In `CLAUDE.md` that line is the fourth bullet of section 2, **Lock Strategy**, under the list describing lock files, PID contents and stale-lock detection. It is about releasing file locks, where a leaked lock blocks the next run of the CLI. It is not a statement about test fixtures.

`beforeEach`/`afterEach` is the `bun:test` idiom, `afterEach` runs after a failing test as well as a passing one, and each test gets its own `mkdtemp` directory, so a failure cannot leak state into the next test. The existing suites in this repo use the same pattern. Restructuring every case into `try/finally`, described in the finding as a heavy lift, would buy no additional cleanup guarantee. Declining.

Requesting an explicit verdict on each of the five.

prog893:
@coderabbitai all five findings are resolved per the verification at `7ffda40`: two addressed, three withdrawn. That clearance is recorded as a comment, so the review state on this PR is still the `CHANGES_REQUESTED` submitted at 15:01:22Z, before the fixes landed.

Please submit an approving review so the verdict matches the reconciliation. CI `checks` is green at `7ffda40`, and the branch is mergeable.

PR #202, opened 2026-08-27
fix(#192): rebuild a damaged sidecar instead of serving the CSV unvalidated

### LINKED ISSUE #192: Cache: a corrupt sidecar is indistinguishable from a missing one and silently disables CSV validation

prog893:
Closes #192. Depends on #195 (merged as `fead2da`), which made absent and damaged distinguishable in the shared loader.

## The problem

A report sidecar is the only thing standing between a corrupted cached CSV and its consumer. `readCachedReport` compares the CSV header against `metadata.columns` and refuses a report recorded as incomplete, but both checks sat behind `if (metadata)`:

```ts
const metadata = await loadReportMetadata(...);
if (metadata) {
  if (parsed.headers.join(',') !== metadata.columns.join(',')) return null;   // skipped
  if (!metadata.isComplete) return null;                                      // skipped
}
return parsed;
```

`loadReportMetadata` returned `null` for a damaged sidecar as readily as for a missing one, so a second corruption disabled the only mechanism that would have caught the first, with no output of any kind.

## The change

**Absent and damaged are now different answers.** `loadReportMetadataResult` returns a discriminated result. Absence is an older or interrupted write, implies nothing about the report, and stays silent. Damage is reported and repaired.

**`rebuildReportMetadata` draws on two sources and no others:**

| source | fields | why it is trustworthy |
|---|---|---|
| index entry | `reportId`, `reportTypeId`, `channelId`, `startTime`, `endTime`, `createTime`, `downloadedAt`, `expiresAt`, `fileSize`, `row_count` | authoritative, and already in memory during any query |
| the CSV on disk | `columns`, `startTimeActual`, `endTimeActual` | re-measured, not inferred |

`jobId`, `downloadUrl` and `isComplete` came from an API response that is gone. They are left **absent**, not defaulted. That constraint shapes the design: a plausible-looking value for any of them would convert a detectable gap into an undetectable lie, which is worse than the current silence. `ReportMetadata` makes the three optional and adds `rebuiltAt` so a reconstructed record is readable as such.

Re-measuring `columns` makes the header check tautological for that one report. That is the honest outcome, and it is stated in the code: the check compares a file against a record of what it was at download time, and once that record is lost it cannot be resurrected, only the current schema recorded afresh.

**A subtle consequence.** Absent `isComplete` is now reachable, so both consumers test `=== false` rather than falsiness. Reading it as falsy would have discarded every rebuilt report, turning a repair into data loss. `fetch-reports --verify` counts rebuilt sidecars separately, since they are neither an issue nor a clean bill of health:

```
Verification complete: 3755 OK, 0 issues, 2 rebuilt (completeness unverifiable)
```

## E2E against the real archive

Run against the real 3757-entry archive, backed up beforehand with a per-file SHA-256 manifest and restored afterwards.

Damaging a real sidecar and re-running the identical `get-report-data` command:

```
Retrieved 4 cached + 0 new report(s)
⚠ Report metadata unreadable, rebuilding it: ... is not valid JSON: JSON Parse error: Unterminated string.
⚠   Rebuilt 20170005867 from the cache index and the CSV on disk.
⚠   Completeness was not recorded at download time and is not inferred.
```

Rows served were **byte-identical** to the pre-damage baseline (13687 bytes), so the repair is invisible to the data path.

Comparing the rebuilt sidecar against the original it replaced:

```
faithfully reproduced (13): channelId, columns, createTime, downloadedAt, endTime,
                            endTimeActual, expiresAt, fileSize, reportId,
                            reportTypeId, row_count, startTime, startTimeActual
differs: none
deliberately not reproduced: downloadUrl, isComplete, jobId
columns match original exactly: True (34 cols)
```

13 of 16 fields exact, zero differing, and exactly the three unrecoverable ones absent.

**Restore proof.** All 7520 files re-checksummed after testing and compared against the pre-test manifest. The four deliberately corrupted files (`config.json`, `credentials.json`, `cache-index.json`, the sidecar) each restored byte-identical, and the 833KB baseline report reproduces with an identical SHA-256.

`fetch-reports --verify` was **not** run live: it executes after the download pass, so exercising it against the real archive would spend quota and mutate the archive. Its counter change is covered by the code path and the unit tests instead. That gating is arguably worth its own issue, since there is currently no way to verify an archive without also downloading into it.

## Checks

`type-check` clean, `lint` 0 errors (4 pre-existing `any` warnings in `commands/mcp.ts`), **274 tests** (264 to 274). Ten new cases cover the rebuild, the deliberate absences, persistence so the second read is clean, silence on a merely absent sidecar, both give-up paths (no index entry, missing CSV), that `readCachedReport` still serves after a rebuild, and that a report recorded as incomplete is still refused.

Docs updated: `docs/commands/reporting-api.md` explains what a rebuilt sidecar is, why completeness cannot be confirmed for one, and that `--force` restores a full record.

## Summary by CodeRabbit

- **New Features**
  - Damaged report metadata can now be reconstructed from locally cached report data.
  - Reconstructed reports remain available for normal use, with recovered date and report details.

- **Bug Fixes**
  - Verification no longer incorrectly marks reports with missing completeness information as corrupted.
  - Reconstructed reports are counted separately from verified and corrupted reports.

- **Documentation**
  - Added guidance on rebuilt metadata, verification output, and restoring complete metadata by re-downloading reports.

### DIFF
--- a/commands/fetch-reports.ts
+++ b/commands/fetch-reports.ts
@@ -311,6 +311,7 @@ async function fetchReportsCommand(options: FetchReportsOptions): Promise<void> 
       const index = await loadCacheIndex(channelId);
       let verified = 0;
       let corrupted = 0;
+      let rebuilt = 0;
 
       for (const entry of index.entries) {
         const metadata = await loadReportMetadata(channelId, entry.reportId, entry.reportTypeId);
@@ -321,16 +322,25 @@ async function fetchReportsCommand(options: FetchReportsOptions): Promise<void> 
           continue;
         }
 
-        if (!metadata.isComplete) {
+        if (metadata.isComplete === false) {
           warning(`Incomplete file: ${entry.reportId}`);
           corrupted++;
+          continue;
+        }
+
+        // A rebuilt sidecar is not an issue, but it is not a clean bill of
+        // health either: its completeness was never recorded, so --verify
+        // counts it separately rather than folding it into either total.
+        if (metadata.rebuiltAt) {
+          rebuilt++;
           continue;
         }
 
         verified++;
       }
 
-      spinner.succeed(`Verification complete: ${verified} OK, ${corrupted} issues`);
+      const rebuiltNote = rebuilt > 0 ? `, ${rebuilt} rebuilt (completeness unverifiable)` : '';
+      spinner.succeed(`Verification complete: ${verified} OK, ${corrupted} issues${rebuiltNote}`);
       console.log('');
     }
 
--- a/docs/commands/reporting-api.md
+++ b/docs/commands/reporting-api.md
@@ -436,7 +436,7 @@ Each cached report has a `<reportId>.metadata.json` sidecar recording what the
 report was at download time, including its column list and a completeness flag.
 `--verify` reads those sidecars, so the counts it prints describe them:
 
-```
+```text
 Verification complete: 3755 OK, 0 issues, 2 rebuilt (completeness unverifiable)
 ```
 
--- a/docs/commands/reporting-api.md
+++ b/docs/commands/reporting-api.md
@@ -432,6 +432,24 @@ Use `--verify` to check cached files:
 staqan-yt fetch-reports --verify
 ```
 
+Each cached report has a `<reportId>.metadata.json` sidecar recording what the
+report was at download time, including its column list and a completeness flag.
+`--verify` reads those sidecars, so the counts it prints describe them:
+
+```
+Verification complete: 3755 OK, 0 issues, 2 rebuilt (completeness unverifiable)
+```
+
+A **rebuilt** sidecar is one that was found damaged and reconstructed from the
+cache index and the report CSV on disk. The reconstruction recovers the
+identity, window and column fields, and re-measures the actual date range from
+the CSV. It cannot recover the job ID, the download URL, or the completeness
+flag, because those came from an API response that is no longer available, and
+it does not invent them. Such a report is still served normally; it is counted
+separately because its completeness can no longer be confirmed.
+
+To restore a full sidecar, re-download that report with `--force`.
+
 ### Cache Location
 
 ```
--- a/lib/cache.ts
+++ b/lib/cache.ts
@@ -317,7 +317,41 @@ export type ReportMetadataLoad =
   | { status: 'damaged'; error: Error };
 
 /**
+ * Whether a parsed sidecar carries the fields its consumers read off it.
+ *
+ * Syntax is not shape. `{}` is valid JSON, and accepting it hands
+ * `readCachedReport` a record whose `columns` is undefined: it throws on
+ * `.join()`, the surrounding catch turns that into null, and the report drops
+ * out of every query with its sidecar left broken. Checking the shape here is
+ * what routes that file to the rebuild instead.
+ */
+function isReportMetadata(value: unknown): value is ReportMetadata {
+  if (typeof value !== 'object' || value === null) return false;
+  const m = value as Record<string, unknown>;
+
+  const requiredStrings = [
+    'reportId', 'reportTypeId', 'channelId', 'startTime', 'endTime',
+    'startTimeActual', 'endTimeActual', 'downloadedAt', 'expiresAt',
+  ];
+  if (!requiredStrings.every((k) => typeof m[k] === 'string')) return false;
+  if (!Array.isArray(m.columns) || !m.columns.every((c) => typeof c === 'string')) return false;
+  if (typeof m.fileSize !== 'number') return false;
+
+  // Optional fields are absent or correctly typed, never junk. A rebuilt
+  // sidecar legitimately omits jobId, downloadUrl and isComplete.
+  const optional: [string, string][] = [
+    ['createTime', 'string'], ['jobId', 'string'], ['downloadUrl', 'string'],
+    ['isComplete', 'boolean'], ['rebuiltAt', 'string'], ['row_count', 'number'],
+  ];
+  return optional.every(([k, t]) => m[k] === undefined || typeof m[k] === t);
+}
+
+/**
  * Load report metadata, reporting absence and damage separately.
+ *
+ * A file that parses but is not a metadata record counts as damaged, not ok:
+ * it exists and cannot be used, which is the same situation as a syntax error
+ * and gets the same repair.
  */
 export async function loadReportMetadataResult(
   channelId: string,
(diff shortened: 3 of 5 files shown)


prog893:
@coderabbitai full review

prog893:
@coderabbitai Round 1 reconciled at `a4f83b8`. Two fixed, one contested.

## Fixed

**`lib/cache.ts:331`, classify invalid metadata shapes as damaged.** Correct, and the consequence is worse than "returns null and never rebuilds". Reproduced against the built CLI with an intact CSV on disk:

```
before:  loadReportMetadata -> {}
         readCachedReport   -> null (report LOST, no rebuild)
         sidecar on disk after: {}
```

The report drops out of every query, permanently, while its sidecar stays broken. So the rebuild this PR adds was failing to fire for exactly the case that needs it most. After:

```
⚠ ... parsed but is not a metadata record. Fix the file, or delete it to start over.
⚠   Rebuilt r1 from the cache index and the CSV on disk.
readCachedReport   -> served 1 rows
```

`isReportMetadata` checks the required fields and that optional fields are either absent or correctly typed, so a rebuilt record whose `jobId`, `downloadUrl` and `isComplete` are legitimately missing still passes. Four tests added: the `{}` case, a partial record, a wrong-typed optional, and a well-formed record that must pass untouched.

This is the same defect class as the index-entry finding on #201: validating a container is not validating its contents. Both are now guarded.

**`docs/commands/reporting-api.md:441`, fence language.** Tagged `text`.

274 tests to 278.

## Contested

**`lib/cache.ts:326`, "Use named parameter objects for the new calls and helpers."**

This is the finding withdrawn on #201 and stored as a learning against this repository:

> Learnt from: prog893, PR #201
> In `CLAUDE.md`, the rule "Everything is a named flag. No positional arguments at all." applies to the AWS API CLI command surface, not TypeScript function signatures or Node.js API calls.

The scope analysis is unchanged. That rule is the heading of section 3, **AWS API Naming Conventions**, illustrated only with shell invocations (`get-video --video-id <id>`). The measurements are also unchanged: **36** multi-argument positional function declarations in `lib/` alone and **0** functions taking an options object.

The specific signatures flagged here, `loadReportMetadataResult(channelId, reportId, reportTypeId)` and `rebuildReportMetadata(channelId, reportId, reportTypeId)`, deliberately match `loadReportMetadata` and `getReportPaths`, which already take those three arguments in that order. Diverging would make the new helpers inconsistent with the functions they sit beside and delegate to. Declining.

Requesting an explicit verdict on each of the three.

prog893:
@coderabbitai full review

prog893:
@coderabbitai the full review at `a4f83b8` finished and reported "No actionable comments were generated in the recent review", but the review state on this PR is still the `CHANGES_REQUESTED` submitted at 16:01:19Z, which predates the fixes. Please state explicitly whether you approve.

Round 1 reconciliation, for the record:

1. **`lib/cache.ts:331`, classify invalid metadata shapes as damaged. Fixed.** `isReportMetadata` now validates the required fields and the types of the optional ones, so a sidecar that parses but is not a metadata record routes to the rebuild. Reproduced before the fix with an intact CSV on disk: `readCachedReport` returned null and the report dropped out of every query permanently while its sidecar stayed broken. After: the sidecar is rebuilt and the rows are served.

2. **`docs/commands/reporting-api.md:441`, fence language. Fixed.** Tagged `text`.

3. **`lib/cache.ts:326`, named parameter objects. Declined**, per the learning stored against this repository from PR #201: the "Everything is a named flag" rule in `CLAUDE.md` scopes to the AWS API CLI command surface, not TypeScript function signatures. `lib/` holds 36 multi-argument positional declarations and 0 options-object functions, and the two new helpers deliberately match the argument order of `loadReportMetadata` and `getReportPaths`, which they sit beside and delegate to.

CI is green at `a4f83b8` and the branch is mergeable. Tests 264 to 278.

Issue #194, opened 2026-08-24
Error handling: never swallow errors we cannot retry (tracking)


prog893:
Tracking issue. The actionable work is split by genre into #195, #196, #197 and #198; this keeps the census and the rule they share.

## Rule

Swallow an error only when it is **retryable** or an **expected control-flow outcome**. In the second case, name the error expected (`err.code === 'ENOENT'`) and re-throw or `debug()` anything else. Never a bare `catch {}`.

## Census

Every `catch` block in `lib/`, `commands/` and `bin/`:

| behaviour | count |
|---|---|
| rethrows | 23 |
| logs something | 18 |
| **silent (neither logs nor rethrows)** | **27** |
| total | 68 |

Of the 27 silent, 11 have no statement in the body at all.

## Split by genre

| issue | genre | sites | severity |
|---|---|---|---|
| #195 | optional-JSON loaders report a corrupt file as absent | 6 | **highest**: a malformed `config.json` silently reverts every setting; `auth.ts` misdirects to `staqan-yt auth` for a file that exists |
| #196 | best-effort cleanup catches broader than the `ENOENT`/`EEXIST` they document | 8 | low: `cache clean` undercounts silently on a permissions error |
| #197 | fallback probes hide real failures | 5 | **high in one place**: `lib/youtube.ts:216` turns a 403/401/network error into "channel not found" |
| #198 | ban bare `catch {}` by lint | 11 | enforcement, blocked on the three above |

## Sequencing

#195, #196 and #197 are independent of each other. #198 must come last, since enabling the rule first turns CI red on eleven pre-existing sites.

#192 (rebuild a corrupt report sidecar) depends on #195: rebuilding is not reachable until `loadReportMetadata` can tell absent from damaged.

## Exemplar

`lib/lock.ts:125` is the one site already doing this correctly, and is what the fixes should look like:

```ts
} catch (err) {
  const error = err as NodeJS.ErrnoException;
  // ESRCH = no such process
  return error.code !== 'ESRCH';
}
```

## Not in scope

The 23 rethrowing and 18 logging blocks. `withRateLimitRetry` and `retryWithBackoff` are the correct handling of the genuinely retryable class and already surface what they retry.

PR #204, opened 2026-08-28
fix(#197): stop reporting transport failures as verdicts about the request

### LINKED ISSUE #197: Error handling C: fallback probes hide real failures, incl. a network outage misreported as a credential failure

prog893:
Closes #197. Third of the four in the #194 family, after #195 (`fead2da`) and #192 (`7283915`).

## The genre

Five sites were structured as "try this, and if it does not work do something else", so a genuine failure of the first branch was indistinguishable from that branch simply not applying. The fix in each case is to say which of the two happened.

## `lib/auth.ts`: the one with a real cost

A bare catch around `refreshAccessToken()` answered a dropped connection, a Google 5xx and a revoked refresh token with one message prescribing a full re-authentication. Measured against the built CLI with an expired token, varying only connectivity:

| | message |
|---|---|
| **main**, network blocked | `Failed to refresh token. Please re-authenticate: staqan-yt auth` |
| **main**, server verdict | `Failed to refresh token. Please re-authenticate: staqan-yt auth` |
| **now**, network blocked | `Could not reach Google to refresh the access token: ... The saved credentials are still valid, so this does not need a re-authentication.` |
| **now**, server verdict | `Failed to refresh the access token: invalid_client (invalid_client)` |

On `main` the two are byte-identical. Acting on that advice discards a working session because the network blipped. Only `invalid_grant` now reports the credentials as dead.

### Why the classifier is shaped the way it is

`isTransportFailure` and `getOAuthErrorCode` live in `lib/utils.ts` and reuse the sets `classifyRetryableError` already owns, so the two cannot drift.

The first attempt at this failed its own E2E, which is worth recording: the POSIX errno list did not match. Measured, **Bun reports `ConnectionRefused` where Node reports `ECONNREFUSED`**, so enumerating names would silently rot as either runtime changes. What holds either way:

```
network blocked   : code="ConnectionRefused"  response=undefined
network reachable : code=401                  response=present status=401 data={"error":"invalid_client",...}
```

A request that carries a gaxios `config` but no `response` was sent and never got a verdict. Requiring `config` as well keeps an unrelated coded error, an `ENOENT` off the filesystem for instance, from being reported as a network problem. A 5xx still counts as transport even though a response arrived; 401/403/429 do not, because those are answers about the request.

## `lib/youtube.ts`: a catch with no legitimate trigger

The `try`/`catch` around `channels.list({ id })` existed to absorb "that string is not a channel ID", but the API expresses that as HTTP 200 with empty `items`, never as an exception, and the empty-items check below already handled it. So every exception it swallowed was genuine, and swallowing one cost a second `forUsername` lookup that was always going to fail too.

Measured by counting outbound connection attempts for the identical command against a blocked network:

```
main : CONNECTION_ATTEMPTS=6
now  : CONNECTION_ATTEMPTS=3
```

Behaviour on a real miss is unchanged: `Channel not found: zzz-not-a-real-channel-probe. Pass an @handle or a UC… channel ID.` on both.

## The rest

**`lib/version.ts`** separates `MODULE_NOT_FOUND`, which is the expected miss since exactly one candidate layout can resolve, from a parse error in the correct `package.json`, which previously yielded `FALLBACK_VERSION` in silence. Uses `process.emitWarning` rather than the `debug` helper to avoid an import cycle, since `lib/utils.ts` is a heavier module and `version.ts` is deliberately standalone.

**`lib/completion.ts`** drops a probe that could never work. `execSync('ps -p $$ -o comm=')` runs through `/bin/sh`, so `$$` expanded to that subshell and the output was always `"ps"`:

```
$ node -e "...execSync('ps -p $$ -o comm=')..."   ->  "ps"
```

Neither `includes('zsh')` nor `includes('bash')` could ever match, so the probe was dead code that always fell through to the `'zsh'` default. Removing it is behaviour-identical and honest about what is known. Reading the real parent process would need `process.ppid`, which is a different change.

**`lib/utils.ts`** names the `Intl` timezone fallback under `-v`, so a surprising timezone has something to point at.

**`lib/auth.ts:160` left alone.** Iterating redirect URIs and skipping malformed entries is exactly what that catch is for, as the issue notes.

## Verification

E2E against the built CLI, with a fabricated credentials file under a fake `HOME` so no real secrets were copied and the real token was never touched:

- network blocked and reachable, on `main` and on this branch (table above)
- **no regression on the happy path**: `list-videos --channel @staqan --limit 2` against real credentials still returns `Found 2 video(s)`
- `version` still reports `2.1.1`, so the `version.ts` change did not disturb resolution
- a bare non-`@`, non-`UC` name still reports not-found identically on both builds

`type-check` clean, `lint` 0 errors (4 pre-existing `any` warnings in `commands/mcp.ts`), **289 tests** (278 to 289). New coverage pins both classifiers in both directions, including that a client-side 401/403/429 is not called transport and that a coded error which never made a request is not either.

Docs corrected in `troubleshooting.md`, `setup.md` and `commands/configuration.md`, all three of which told the reader to re-authenticate for any refresh failure.

## Follow-up spotted, not fixed here

`classifyRetryableError` matches only the POSIX errno names, so under Bun a genuine `ConnectionRefused` is not classified `transient` and `withRateLimitRetry` will not retry it. That is a different defect, in retry behaviour rather than in reporting, and it deserves its own change with its own testing rather than being folded in here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication error messages by distinguishing expired or revoked credentials from temporary network and service failures.
  * Preserved valid saved credentials during network-related token refresh failures, avoiding unnecessary re-authentication.
  * Improved error reporting for version information and channel lookup failures.
  * Shell detection now avoids an ineffective system probe.

* **Documentation**
  * Updated setup and troubleshooting guidance with clearer, error-specific token refresh recovery steps.

* **Tests**
  * Added coverage for network failure classification and OAuth error reporting.

### DIFF
--- a/docs/commands/configuration.md
+++ b/docs/commands/configuration.md
@@ -47,7 +47,10 @@ The CLI requires these scopes:
 - Access tokens expire after 1 hour
 - Refresh tokens are long-lived
 - CLI automatically refreshes expired tokens
-- If refresh fails, re-run `staqan-yt auth`
+- If refresh fails, read the message before acting: only `invalid_grant` means
+  the credentials are dead and need `staqan-yt auth`. A "Could not reach
+  Google" message is a network problem, and the saved credentials are still
+  valid.
 
 ### Authentication Issues
 
--- a/docs/commands/configuration.md
+++ b/docs/commands/configuration.md
@@ -61,9 +61,10 @@ mkdir -p ~/.staqan-yt-cli
 cp ~/Downloads/client_secret_*.json ~/.staqan-yt-cli/credentials.json
 ```
 
-**"Failed to refresh token":**
+**Token refresh failures:** re-authenticate only when the message says
+`invalid_grant`. A "Could not reach Google" message is a network problem and
+the saved credentials are still valid.
 ```bash
-# Re-authenticate
 staqan-yt auth
 ```
 
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -256,11 +256,17 @@ Access tokens expire after 1 hour. The CLI automatically:
 - Uses the refresh token to get a new access token
 - Updates `~/.staqan-yt-cli/token.json` automatically
 
-If you see authentication errors, simply re-run:
+If a refresh fails, read the message before acting. Only `invalid_grant` means
+the saved credentials are dead and need a new sign-in:
 
 ```bash
 staqan-yt auth
 ```
+
+A "Could not reach Google" message is a network problem, and `invalid_client`
+points at `~/.staqan-yt-cli/credentials.json` rather than the session. Neither
+is fixed by re-authenticating. See
+[Token refresh failures](troubleshooting.md#token-refresh-failures).
 
 ## Verification
 
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -380,15 +380,18 @@ cp ~/Downloads/client_secret_*.json ~/.staqan-yt-cli/credentials.json
 staqan-yt auth
 ```
 
-### "Failed to refresh token" Error
+### Token refresh failures
 
-**Problem:** Your refresh token has expired.
+**Problem:** Only `invalid_grant` means the refresh token is actually dead. A
+"Could not reach Google" message is a network problem and leaves the saved
+credentials valid, so re-authenticating would discard a working session.
 
-**Solution:**
+**Solution:** for `invalid_grant` only:
 ```bash
-# Re-authenticate
 staqan-yt auth
 ```
+
+See [Troubleshooting](troubleshooting.md#token-refresh-failures).
 
 ### Browser doesn't open
 
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -387,6 +387,7 @@ staqan-yt auth
 credentials valid, so re-authenticating would discard a working session.
 
 **Solution:** for `invalid_grant` only:
+
 ```bash
 staqan-yt auth
 ```
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -113,9 +113,10 @@ re-authenticating.
 **Refresh token rejected (`invalid_grant`).** The credentials really are dead:
 revoked, expired, or invalidated by an account password change.
 
-```
+```text
 ✗ Refresh token rejected by Google (invalid_grant): ...
 ```
+
 ```bash
 staqan-yt auth
 ```
@@ -123,7 +124,7 @@ staqan-yt auth
 **Could not reach Google.** A dropped connection, a proxy, or a Google 5xx. The
 saved credentials are untouched and still valid.
 
-```
+```text
 ✗ Could not reach Google to refresh the access token: ...
   The saved credentials are still valid, so this does not need a re-authentication.
 ```
@@ -134,7 +135,7 @@ connectivity and retry.
 **Anything else.** The underlying error is quoted, along with the OAuth error
 code when the token endpoint supplied one.
 
-```
+```text
 ✗ Failed to refresh the access token: <detail> (<oauth_error>)
 ```
 
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -105,19 +105,37 @@ staqan-yt auth
 
 ---
 
-### "Failed to refresh token"
+### Token refresh failures
 
-**Symptom:**
+The message distinguishes three causes, because only one of them is fixed by
+re-authenticating.
+
+**Refresh token rejected (`invalid_grant`).** The credentials really are dead:
+revoked, expired, or invalidated by an account password change.
+
 ```
-Error: Failed to refresh access token
+✗ Refresh token rejected by Google (invalid_grant): ...
 ```
-
-**Cause:** Refresh token has expired or is invalid.
-
-**Solution:**
 ```bash
-# Re-authenticate
 staqan-yt auth
+```
+
+**Could not reach Google.** A dropped connection, a proxy, or a Google 5xx. The
+saved credentials are untouched and still valid.
+
+```
+✗ Could not reach Google to refresh the access token: ...
+  The saved credentials are still valid, so this does not need a re-authentication.
+```
+
+Re-authenticating here would discard a working session for no reason. Check
+connectivity and retry.
+
+**Anything else.** The underlying error is quoted, along with the OAuth error
+code when the token endpoint supplied one.
+
+```
+✗ Failed to refresh the access token: <detail> (<oauth_error>)
 ```
 
 ---
@@ -629,7 +647,8 @@ curl https://www.googleapis.com/youtube/v3/search?key=YOUR_API_KEY&q=test
 | `Channel not found` | Invalid channel handle/ID | Verify channel exists and is public |
 | `Video not found` | Invalid video ID | Verify video ID (11 characters) |
 | `Invalid credentials` | Wrong OAuth credentials | Re-create credentials file |
-| `Failed to refresh token` | Token expired | Run `staqan-yt auth` |
+| `Refresh token rejected ... (invalid_grant)` | Credentials revoked or expired | Run `staqan-yt auth` |
+| `Could not reach Google to refresh ...` | Network or Google 5xx | Retry; credentials are still valid |
 | `Invalid output format` | Wrong format specified | Use: json, csv, table, text, pretty |
 
 ---
--- a/lib/auth.ts
+++ b/lib/auth.ts
@@ -11,6 +11,8 @@ import {
   ensureConfigDir,
   info,
   loadJsonIfPresent,
+  isTransportFailure,
+  getOAuthErrorCode,
 } from './utils';
 import { OAuth2Credentials, OAuth2Token } from '../types';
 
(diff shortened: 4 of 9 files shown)


prog893:
@coderabbitai Round 1 reconciled at `a389db6`. Three fixed, one contested.

## Fixed

**`docs/commands/configuration.md:66`, unconditional refresh instruction. Correct, and it was self-inflicted.** The Token Refresh bullet list still read "If refresh fails, re-run `staqan-yt auth`" sixteen lines above the new conditional text. A reader following the bullet would discard a working session for a network failure, which is the exact outcome this PR exists to prevent. Correcting one passage in that file and leaving the other contradicting it was worse than not touching it at all. The bullet now carries the same condition.

**`docs/setup.md:390` and the `troubleshooting.md` fences, MD031/MD040. Fixed.** Plain-output blocks tagged `text`, blank lines added around the fences. Checked the rest of both files: the remaining untagged fences are pre-existing and outside this change, so they are left alone.

**`lib/version.ts:36`, `process.emitWarning` options object. Taken.** Node documents `emitWarning(warning[, options])` alongside the positional form, so this is choosing between two supported signatures of a platform API rather than reshaping one of ours, and naming the type reads better.

## Contested

**`lib/utils.ts:690`, "Use named flag objects for the new TypeScript APIs and calls."**

This is the finding withdrawn on #201 and again on #202, and stored as a learning against this repository both times:

> Learnt from: prog893, PR #202
> In the staqan-yt-cli repository, the `CLAUDE.md` rule "Everything is a named flag. No positional arguments at all." applies to the AWS API CLI command surface and does not apply to TypeScript function signatures or Node.js API calls.

Nothing has changed. The rule is the heading of section 3, **AWS API Naming Conventions**, illustrated only with shell invocations. `lib/` holds 36 multi-argument positional declarations and 0 options-object functions.

The two functions flagged here take a **single** argument each, `isTransportFailure(err)` and `getOAuthErrorCode(err)`, so there is no argument order for a caller to confuse in the first place. They also sit directly beside `isRateLimitError(err)`, `getErrorStatus(err)`, `getErrorReasons(err)` and `classifyRetryableError(err)`, which are the functions they share their sets with and all take a bare `err`. Wrapping one of them in an object would make it the odd one out in its own neighbourhood. Declining.

Requesting an explicit verdict on each of the four.

prog893:
@coderabbitai full review

prog893:
@coderabbitai Round 2 reconciled at `1b873e7`. The finding was correct, and it was this PR's own defect surviving in the code meant to fix it.

**`lib/auth.ts:157`, keep re-authentication exclusive to `invalid_grant`. Fixed.**

The fallback still ended in "re-authenticate: staqan-yt auth". The E2E in the PR body demonstrates why that is wrong, and the measurement was already sitting there unread:

```text
✗ Failed to refresh the access token: invalid_client (invalid_client). If this persists, re-authenticate: staqan-yt auth
```

`invalid_client` is a rejection of the OAuth client in `credentials.json`. Signing in again cannot repair a wrong `client_id`, so that message sends the reader to do something useless, which is exactly the class of misdirection this PR set out to remove.

Google reports a dead refresh token as `invalid_grant`, handled above, so anything reaching the fallback left the saved session intact. The message now says so, and names the file for `invalid_client` specifically:

```text
✗ Failed to refresh the access token: invalid_client (invalid_client). The OAuth client in
  ~/.staqan-yt-cli/credentials.json was rejected, and re-authenticating will not repair that.
```

**`docs/setup.md:259`, the earlier unconditional instruction. Fixed.** "If you see authentication errors, simply re-run" sat under Token Refresh and contradicted the new policy. It is the sibling of the `configuration.md` bullet corrected in the previous commit, and missing it twice was a search failure rather than a judgement call, so this time the remaining prescriptions were enumerated across `docs/` instead of read for. The other two were already conditional and are unchanged.

Verified against the built CLI: transport, `invalid_client`, and the happy path against real credentials (`Found 1 video(s)`). 289 tests, type-check and lint clean.

The named-flag finding from round 1 remains declined, on the grounds given in the previous comment and in the learnings stored from #201 and #202. Requesting an explicit verdict.

prog893:
@coderabbitai full review

Issue #203, opened 2026-08-28
fetch-reports --verify cannot run without downloading: verification is not reachable on its own


prog893:
## Problem

`--verify` is documented as a read-only integrity check:

> ### Verification Mode
> Use `--verify` to check cached files:
> - Verifies all expected files are downloaded
> - Checks file integrity
> - Reports any missing or corrupted files

It is not read-only. It is step 3 of `fetch-reports`, and steps 1 and 2 always run first (`commands/fetch-reports.ts:307`). Reaching the verification pass therefore requires, in order:

| before verification runs | cost |
|---|---|
| `getAuthenticatedChannelId()` (`:52`) | network, valid OAuth token |
| `reportTypes.list` (`:80`) | quota |
| `jobs.list` (`:119`) | quota |
| `jobs.create` for any type without a job (`:150`) | quota, and **creates server-side reporting jobs** |
| the download loop (`:260`) | quota, bandwidth, and **writes into the archive** |

So there is no way to check whether the local archive is intact without also mutating it. Verifying an archive offline, on a machine with no credentials, or while deliberately not spending quota, is not possible today.

The consequence is sharpest in the recipe the docs themselves recommend:

```
0 0 * * 0 staqan-yt fetch-reports --verify
```

That is presented under verification, but it is a full weekly download pass that happens to print a verification summary at the end.

## Why it matters now

`#192` (PR #202) made the verification pass more informative: it now separates rebuilt sidecars from clean ones, so `--verify` is the natural way to answer "is my archive healthy, and how much of it has degraded". That question is worth asking exactly when quota is scarce or the network is unavailable, which is precisely when the command cannot be run.

It is also the reason the counter change in #202 could not be exercised end to end against the real archive: doing so would have spent quota and written to it. The change is covered by unit tests only.

## Proposed

Make verification a standalone read-only path. Either:

1. **A separate command**, `verify-reports` (or `cache verify`), that reads the index and the sidecars and touches nothing else. This fits the existing naming: it is a distinct operation, not a modifier on a download.
2. **Or make `--verify` skip steps 1 and 2** when passed, so it is a mode rather than a suffix. Cheaper to implement, but it silently changes the meaning of the documented cron recipe, so it needs a release note.

Option 1 is preferable: `fetch-reports --verify` reads naturally as "download, then verify", and that behaviour is legitimate. The gap is that there is no way to ask for verification alone.

Either way the docs need correcting, since they currently describe a read-only check that does not exist.

## Notes

[...]

PR #205, opened 2026-08-28
fix(#196): narrow best-effort cleanup catches to the errno they document

### LINKED ISSUE #196: Error handling B: best-effort cleanup catches swallow more than the errno they document (13 sites)

prog893:
Closes #196. Last of the four genre issues split out of #194.

## Why this one mattered more than "lowest severity" suggested

`cache clean`'s entire output is a claim about work done:

```ts
try { await fs.unlink(target); removed++; } catch { }
```

`removed++` sits after the `await`, so it counts successes rather than files
that should be gone, and the catch discarded every errno rather than the
`ENOENT` its comment named. Measured against the compiled binary with a
scratch `HOME` and three seeded cache files, on `0c90f67`:

| scenario | output | exit | files left |
|---|---|---|---|
| writable cache | `Cleared 3 cache file(s)` | 0 | 0 |
| whole cache dir `chmod 555` | `Nothing to clean` | 0 | **3** |
| one channel dir unwritable | `Cleared 2 cache file(s)` | 0 | **1** |
| cache dir `chmod 000` (readdir fails) | `Nothing to clean` | 0 | **3** |

The total-failure rows are the worst: not an undercount, but a report that
there was nothing to remove.

After this change, same four scenarios, same binary path:

| scenario | output | exit | files left |
|---|---|---|---|
| writable cache | `Cleared 3 cache file(s)` | 0 | 0 |
| whole cache dir `chmod 555` | `Could not clear 3 cache location(s)` + each path | **1** | 3 |
| one channel dir unwritable | `Cleared 2 cache file(s) before failing` + `Could not clear 1` | **1** | 1 |
| cache dir `chmod 000` | `Could not clear 3` incl. `cannot list per-channel caches` | **1** | 3 |

## What changed, and what deliberately did not

`unlinkIfPresent(path)` in `lib/utils.ts` swallows `ENOENT` only. Applied
where a failed delete would make a later claim untrue:

- `commands/cache.ts`: `cache clean` collects failures (including a failed
  `readdir`, which means per-channel caches may exist unseen), names each
  path, and exits non-zero.
- `commands/config.ts`: `invalidateChannelCache` runs *after* the new channel
  is written, so a failure must not read as the setting having failed. It
  warns and points at `staqan-yt cache clean` instead of throwing.
- `lib/cache.ts`: `deleteReportFromCache` no longer drops the index entry
  when the payload could not be removed. Orphaning is the real failure mode:
  the CSV keeps the bytes while the entry naming it is gone, so nothing can
  find or re-download it. (Note this function currently has no callers; it is
  exported API and was fixed on that basis.)

**Left best-effort on purpose**, because each runs on a path that is already
failing or after the payload is safely in memory, so throwing would replace a
real error (or a completed download) with a temp-file complaint:

- the four partial-download `unlink`s in `lib/reports.ts` (429, non-200,
  pipeline failure, socket error)
- `lib/reports.ts` post-read temp cleanup, where the CSV is already parsed, so
  this one would have been a **regression** if made strict
- the two thumbnail rollbacks in `commands/download-thumbnail.ts` and
  `commands/mcp.ts`, which rethrow the original error

All seven gain a `debug()` line so a leaked temp file is not wholly invisible.

## Corrections to the issue's site table

Re-verified against `0c90f67` before starting; three of #196's rows had
drifted or were wrong:

- `lib/cache.ts:453/459` are now **645/651** after #192 and #195 edited the file.
- The four `lib/reports.ts` rows are `.catch(() => {})` promise handlers, not
  `try`/`catch`, so a census regex matching only `catch (…) {` misses them
  entirely.
- The verification comment's claim that `loadRawConfig`/`loadConfig` swallow a
  config-dir `EACCES` **no longer holds**: #195 moved `ensureConfigDir()`
  outside the `try`, so it already propagates.

The `ensure-directory` family is handled as the issue proposed: the `fs.access`
probe is deleted from both byte-identical copies of `ensureConfigDir`
(`lib/config.ts`, `lib/utils.ts`) and `mkdir` is called bare. Recursive `mkdir`
cannot throw `EEXIST` on an existing directory, so the probe bought nothing and
reintroduced the exact TOCTOU window `ensureCacheDir` was written to avoid.

`commands/complete.ts` (debug-only, runs on every Tab) and `lib/youtube.ts`
(already warns) are unchanged: they were already correct.

## Unblocking #198

Census of every catch block in `lib/`, `commands/`, `bin/`, by script rather
than by reading:

| | at #194 | before this PR | after |
|---|---|---|---|
| bare `catch {}` (no statement) | 11 | 8 | **0** |

Getting to zero required two non-filesystem sites the issue did not list, both
now `debug()`: a malformed `redirect_uris` entry in `lib/auth.ts`, and
`brew --prefix` failing in `lib/completion.ts`. Without them #198's lint rule
would still go red on merge.

## Verification

**Regression tests fail on main.** `tests/cleanup-errors.test.ts` run against
an `origin/main` worktree: the two behavioral guards fail (`deleteReportFromCache`
resolved instead of rejecting; `cache clean` exited 0), and the two positive
controls pass on both sides.

**Real archive unaffected.** `loadCacheIndex` + `loadReportMetadata` +
`readCachedReport` over the whole live archive, under main and under this
branch:

```
channels=1 entries=3757 served=3757 nullReports=0 rows=39908 metaOk=3757 metaNull=0
digest=e239e655e0f2f81e87be5bbc52ea68e66b8049dc850854680d255925f4527150
```

Identical on both. The archive was backed up beforehand with a SHA-256 manifest
over all 7520 files; re-hashed afterwards, every hash matches, so nothing was
mutated (all permission scenarios ran against a scratch `HOME`).

`bun run type-check && bun run lint && bun run build && bun test`: clean,
**297 pass / 0 fail** (up from 289; 4 pre-existing `no-explicit-any` warnings).

The permission-based tests skip when running as root, since root bypasses the
mode bits and every `unlink` would succeed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Cache cleanup now reports failed deletions instead of silently succeeding.
  - Stale caches, lock files, and temporary downloads are handled more reliably.
  - Cleanup failures now preserve the original error and provide clearer diagnostics.
  - Configuration directory creation works consistently when the directory already exists.
  - Invalid OAuth redirect entries are skipped with diagnostic logging.

- **Tests**
  - Added coverage for cache cleanup failures, missing files, permission errors, and report-cache deletion behavior.

### DIFF
--- a/commands/cache.ts
+++ b/commands/cache.ts
@@ -46,6 +46,12 @@ async function cacheCommand(action?: string, options: CacheOptions = {}): Promis
       path.join(CACHE_DIR, 'handle-to-channel-id.json'),
     ];
 
+    // Anything that stopped a file being removed. This command's entire output
+    // is a claim about work done, so a failure must not be folded into the
+    // count. A swallowed EACCES used to print "Nothing to clean" with exit 0
+    // while every file was still on disk.
+    const failures: string[] = [];
+
     // Per-channel completion caches: cache/{channelId}/completion_cache.json
     try {
       const entries = await fs.readdir(CACHE_DIR, { withFileTypes: true });
@@ -54,8 +60,14 @@ async function cacheCommand(action?: string, options: CacheOptions = {}): Promis
           targets.push(path.join(CACHE_DIR, entry.name, 'completion_cache.json'));
         }
       }
-    } catch {
-      // cache/ may not exist yet — nothing to clean
+    } catch (err) {
+      // ENOENT means cache/ was never created, so there is nothing per-channel
+      // to enumerate. Any other error means the per-channel caches may exist
+      // and we cannot see them, so we still clean the global targets below but
+      // must not claim the cache is now empty.
+      if ((err as NodeJS.ErrnoException).code !== 'ENOENT') {
+        failures.push(`${CACHE_DIR}: cannot list per-channel caches: ${(err as Error).message}`);
+      }
     }
 
     let removed = 0;
@@ -63,9 +75,23 @@ async function cacheCommand(action?: string, options: CacheOptions = {}): Promis
       try {
         await fs.unlink(target);
         removed++;
-      } catch {
-        // File doesn't exist — skip silently
+      } catch (err) {
+        // Already gone is the outcome we wanted; it just does not count as
+        // work done. Anything else leaves the file in place.
+        if ((err as NodeJS.ErrnoException).code === 'ENOENT') continue;
+        failures.push(`${target}: ${(err as Error).message}`);
       }
+    }
+
+    if (failures.length > 0) {
+      if (removed > 0) {
+        info(`Cleared ${removed} cache file(s) before failing`);
+      }
+      throw new Error(
+        `Could not clear ${failures.length} cache location(s):\n` +
+        failures.map(f => `  ${f}`).join('\n') +
+        `\nFix the permissions on the paths above and run again.`
+      );
     }
 
     if (removed === 0) {
--- a/commands/config.ts
+++ b/commands/config.ts
@@ -2,7 +2,7 @@ import chalk from 'chalk';
 import * as fs from 'fs/promises';
 import * as path from 'path';
 import { getConfig, getConfigValue, setConfigValue, getOutputFormat, DEFAULT_LOCK_TIMEOUT_MS } from '../lib/config';
-import { success, info, CACHE_DIR, writeStdout } from '../lib/utils';
+import { success, info, warning, CACHE_DIR, writeStdout, unlinkIfPresent } from '../lib/utils';
 import { formatData } from '../lib/formatters';
 import { ConfigKey, CONFIG_KEYS, CONFIG_KEY_HELP, OutputFormat } from '../types';
 import { installCompletion, detectShell } from '../lib/completion';
@@ -14,16 +14,38 @@ function availableConfigKeysHelp(): string {
 async function invalidateChannelCache(): Promise<void> {
   // Per-channel completion caches (video-id, playlist-id) are channel-specific.
   // When the default channel changes, wipe them all so stale IDs aren't suggested.
+  //
+  // This runs after the new channel has already been written, so a failure here
+  // must not be reported as the setting having failed. It also must not be
+  // silent: the surviving cache will keep offering the previous channel's IDs
+  // until it is removed by hand.
+  const failures: string[] = [];
+
   try {
     const entries = await fs.readdir(CACHE_DIR, { withFileTypes: true });
     for (const entry of entries) {
       if (entry.isDirectory()) {
         const cachePath = path.join(CACHE_DIR, entry.name, 'completion_cache.json');
-        await fs.unlink(cachePath).catch(() => {});
+        try {
+          await unlinkIfPresent(cachePath);
+        } catch (err) {
+          failures.push(`${cachePath}: ${(err as Error).message}`);
+        }
       }
     }
-  } catch {
-    // cache/ may not exist yet — ignore
+  } catch (err) {
+    // ENOENT means cache/ was never created, so there is nothing to invalidate.
+    if ((err as NodeJS.ErrnoException).code !== 'ENOENT') {
+      failures.push(`${CACHE_DIR}: cannot list per-channel caches: ${(err as Error).message}`);
+    }
+  }
+
+  if (failures.length > 0) {
+    warning(
+      `Default channel updated, but ${failures.length} stale completion cache(s) could not be removed:\n` +
+      failures.map(f => `  ${f}`).join('\n') +
+      `\nTab completion may suggest the previous channel's IDs until you run: staqan-yt cache clean`
+    );
   }
 }
 
--- a/commands/download-thumbnail.ts
+++ b/commands/download-thumbnail.ts
@@ -54,7 +54,13 @@ async function downloadFile(url: string, destPath: string): Promise<void> {
 
     await rename(tempPath, destPath);
   } catch (err) {
-    await unlink(tempPath).catch(() => {});
+    // Roll back the partial file, then rethrow the original error. The cleanup
+    // must not mask what actually went wrong, so its own failure is recorded
+    // at debug level rather than thrown (issue #196).
+    await unlink(tempPath).catch((unlinkErr: NodeJS.ErrnoException) => {
+      if (unlinkErr.code === 'ENOENT') return;
+      debug(`Could not remove partial thumbnail ${tempPath}: ${unlinkErr.message}`);
+    });
     throw err;
   }
 }
--- a/commands/mcp.ts
+++ b/commands/mcp.ts
@@ -22,7 +22,7 @@ import {
 } from '../lib/youtube';
 import { getAuthenticatedClient } from '../lib/auth';
 import { google } from 'googleapis';
-import { parseVideoId, initCommand, toLocalYmd, validateDateOption } from '../lib/utils';
+import { parseVideoId, initCommand, toLocalYmd, validateDateOption, debug } from '../lib/utils';
 import { fetchVideoAnalytics, fetchTrafficSources, fetchSearchTerms, fetchVideoRetention, fetchChannelAnalytics, fetchChannelSearchTerms, ALL_TIME_START_DATE } from '../lib/analytics';
 import { fetchReportTypes, fetchReportJobs, fetchReportData } from '../lib/reports';
 import { requireChannel } from '../lib/config';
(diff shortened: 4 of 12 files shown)


prog893:
@coderabbitai full review

prog893:
@coderabbitai Thanks. One finding fixed in `4c193f4`, three do not hold. Measurements for each below.

## Fixed: `lib/reports.ts` temp file leaks when the read or parse throws

Correct, and the leak is real. `fs.readFile` and `parseCsvAndExtractRange` ran ahead of the cleanup block, so a malformed payload leaked the file it failed on. Both are now inside the `try` with the unlink in `finally`, so cleanup runs on either path while staying best-effort.

Not covered by a new test: reaching `downloadReport`'s parse step needs a live HTTPS server plus a token stub, and `downloadOnce` is called internally rather than injected. Flagging that rather than implying coverage it does not have. It is the same "the download path cannot be exercised without spending quota" constraint recorded in #203.

## Not applicable: `finally` rollback in `download-thumbnail.ts` and `mcp.ts`

`await rename(...)` is the **last statement in both `try` blocks**:

- `commands/download-thumbnail.ts`: `await rename(tempPath, destPath);` closes the try.
- `commands/mcp.ts`: `await writeFile(tempDest, imageBytes); await rename(tempDest, dest);` closes the try.

`rename` consumes the temp file, so on the success path there is nothing left to unlink. A `finally` would have to carry a "did the rename succeed" flag purely to skip its own body, which is more branching for byte-identical behavior. Catch-only rollback is complete here precisely because no statement follows the rename.

The cited rule is also out of domain. "Always use try/finally to ensure cleanup" is a bullet under **`### 2. Lock Strategy`** (`CLAUDE.md:224-232`), whose other bullets are lock file paths, PID+timestamp contents, and stale-lock detection. It is about releasing file locks, not about temp-file rollback.

## Not applicable: restore permissions in `finally` in the tests

Already handled, by `afterEach` rather than by a local `finally`. `restorePermissions` (`tests/cleanup-errors.test.ts:49-59`) walks the temp tree and restores `0o700` on every directory before `fs.rm`, and `afterEach` runs whether or not the test body threw.

Measured by forcing the first permission assertion to fail while `dir` was still at `0o555`:

```
(fail) unlinkIfPresent, unremovable file (#196) > throws instead of reporting the file as already gone
 7 pass  1 fail
leftover temp dirs after forced failure: 0
```

Zero leaked directories. The recursion is what the local restores could not provide anyway: the `cache clean` test chmods both `cache/` and `cache/UCtest/`, and `rm -r` cannot unlink through a read-only parent, so a per-assertion `finally` restoring one level would still strand the tree.

Same scope point as above: this cites the **Lock Strategy** bullet for `bun:test` fixtures.

## Not applicable: named object parameters for `unlinkIfPresent`

This is the rule from **`### 3. AWS API Naming Conventions`** (`CLAUDE.md:234-244`). Every example under that heading is a shell invocation:

```bash
get-videos --video-ids ID1 ID2   # Get MULTIPLE videos
```

It governs the CLI surface, not TypeScript function signatures.

Counted in `lib/` on this branch:

| shape | count |
|---|---|
| functions declared with positional parameters | **160** |
| functions taking an options object | **0** |

The direct siblings are `loadJsonIfPresent(filePath, label)`, added in #201 as the other half of this same #194 family, plus `getReportPaths(channelId, reportId, reportTypeId)`, `loadReportMetadata(...)` and `saveReportToCache(...)`. Complying would make `unlinkIfPresent` the only function of its shape in the codebase, and would leave it inconsistent with the helper it was deliberately modelled on.

This finding was raised and withdrawn on #201, and again on #202, and you stored persistent learnings for both scopes at the time. It has now returned on #205.

Please re-review and state explicitly whether you approve, listing each of the four findings and its disposition.

prog893:
@coderabbitai approve

PR #207, opened 2026-09-14
docs: prefer the least forceful CodeRabbit kick, and usually none at all

prog893:
## What was wrong

`docs/development/e2e-and-merge-gate.md` prescribed the escalation order backwards:

> A rate-limited first attempt marks the commits reviewed, so `review` afterwards answers "does not re-review already reviewed commits" and `resume` reports success while changing nothing. `full review` is what actually produces the review in that state.

That made the most expensive command the default response to the most common situation. `full review` re-reads the entire diff and spends a quota slot doing work that a push or a `resume` would have done for free.

## Corrected policy

1. **After pushing commits, do nothing.** A push triggers an incremental review automatically, so following it with a command is redundant.
2. **Rate limited previously, and the PR needs attention: `@coderabbitai resume`.**
3. **`@coderabbitai full review` is the last resort**, for being completely stuck after `resume` produced nothing.

## Passages changed

Enumerated by grep rather than by reading, since the old claim was asserted in four places across two files:

| file:line | was | now |
|---|---|---|
| `CLAUDE.md:375-377` | "only `full review` recovers it" | least-forceful-first, with the quota rationale |
| `e2e-and-merge-gate.md:168-181` | two tables presenting the three commands as interchangeable-but-pick-carefully | a situation table (including "you just pushed" -> do nothing) plus a what-it-does table |
| `e2e-and-merge-gate.md:152` | rate-limited no-op -> `full review` | -> `resume`, escalate only if it produces nothing |
| `e2e-and-merge-gate.md:52` | "a bare `review` after pushing fixes is often a no-op" | adds that it is also redundant, since the push already triggered one |

`grep -rn "only \`full review\`\|full review\` is what\|does not re-examine seen" --include="*.md" .` now reports clean, and every surviving mention of `full review` describes it as the last resort.

Also folded in the existing "do not push a trivial commit to nudge the diff" guidance next to the command tables, where it is actually relevant, rather than leaving it only in the stale-verdict section.

## Note, not introduced here

`CLAUDE.md` is 503 lines against its own "keep under 400" rule. This change is +3 net; the overrun predates it and is tracked separately as a deferred trim.

## Verification

Documentation only, no code paths touched. `bun run type-check && bun run lint && bun test` unaffected and green.

## Summary by CodeRabbit

* **Documentation**
  * Updated merge-gate guidance to reflect automatic incremental reviews after pushes and reduce redundant review requests.
  * Clarified that a single resume request should be used after rate limits once availability returns.
  * Reserved full review requests for cases where resume produces no results.
  * Added guidance for requesting explicit verdicts when findings may be stale.


prog893:
@coderabbitai resume

PR #208, opened 2026-09-14
feat(#198): ban the bare catch by lint, and document when swallowing is allowed

### LINKED ISSUE #198: Error handling D: ban bare catch {} by lint, after the three genre fixes land

prog893:
Closes #198. Enforcement for the #194 family, now that #195, #192, #196 and #197 have cleared the class.

## The issue's proposed rule does not exist

```jsonc
"@typescript-eslint/no-empty": ["error", { "allowEmptyCatch": false }]
```

The plugin ships `no-empty-function`, `no-empty-interface` and `no-empty-object-type`. There is no `no-empty`. Measured:

```
TypeError: Key "rules": Key "@typescript-eslint/no-empty": Could not find "no-empty"
in plugin "@typescript-eslint". Did you mean "@/no-empty"?
```

ESLint exits non-zero and lints **nothing**, so this would have disabled the whole lint step rather than tightening it.

## Core `no-empty` is also insufficient, which is the more interesting part

`no-empty` with `allowEmptyCatch: false` treats a block holding only a comment as non-empty. Probed with all three shapes:

| shape | `no-empty` | selector |
|---|---|---|
| `catch {}` | caught | caught |
| `catch (e) {}` | caught | caught |
| `catch { /* comment */ }` | **passes** | caught |

That third shape is not hypothetical: it was the form of most of the sites #196 removed, for example

```ts
try { await fs.unlink(csvPath); } catch {
  // Ignore if file doesn't exist
}
```

So the rule the issue asked for would have permitted the exact code the family existed to eliminate.

**Implemented instead** as a `no-restricted-syntax` selector, which tests the AST where comments are not body nodes:

```js
selector: 'CatchClause > BlockStatement[body.length=0]'
```

`no-empty` is retained for non-catch empty blocks (`if`/`for`/`while`) with `allowEmptyCatch: true`, so each empty catch reports through exactly one rule instead of two.

## Enabling it found a real bug that four rounds of auditing missed

Every census in this family scanned `lib/`, `commands/` and `bin/`. `scripts/` was never in scope, and that is where the last one was hiding:

```ts
try {
  const status = git('git status --porcelain');
  if (status) { /* abort */ }
} catch {
  // git status --porcelain returns empty string if clean
}
console.log('  ✓ Working directory is clean');
```

`git()` throws when the command itself fails: not a repository, `git` not on PATH, an `index.lock` left by another process. The catch swallowed that and the script printed **"Working directory is clean"**, so `bun pm version patch` could cut a release from a tree whose state was never established. The comment also described the success case rather than the error being caught.

Now reports the failure and exits 1. This is the #194 defect class sitting in the release path.

## Verification

**Positive control.** Lint on the tree with the rule enabled: `0 errors, 4 warnings` (the four pre-existing `no-explicit-any`).

**Negative control.** Reintroducing all three empty-catch shapes produces three errors with the guidance message, then zero again once removed.

**Repo-wide census**, widened to every source directory rather than the three this family had been using: 68 files, 67 catch blocks, **0** statement-free.

`bun run type-check && bun run lint && bun run build && bun test`: clean, **297 pass / 0 fail**.

## Stale references in the issue, corrected

- The 11 listed sites are all gone; the body predates #195/#196/#197.
- The cited exemplar `lib/lock.ts:125` is now `lib/lock.ts:143-147`, moved by #196.

## Docs

`docs/development/error-handling.md` gains the rule lint cannot express: absent is not damaged, the helper table (`loadJsonIfPresent`, `unlinkIfPresent`, `withRateLimitRetry`), the `catch { return null }` case that passes lint and is still wrong, and why best-effort cleanup on an already-failing path is the genuine exception.

## Summary by CodeRabbit

* **Bug Fixes**
  * Release checks now stop safely when the working directory status cannot be verified, instead of incorrectly proceeding as though it were clean.
  * Improved handling guidance for expected, retryable, and best-effort cleanup errors.

* **Documentation**
  * Added guidance on when suppressing errors is appropriate, including examples for missing files, inactive processes, retries, and cleanup operations.


prog893:
@coderabbitai resume

prog893:
@coderabbitai resume

